PR #5, opened 2022-08-30
skipped making it looking better, looks good enough to me)

tarkaiev:


IvanRamyk (changes requested):
Please, add screenshots to the PR

PR #28, opened 2022-11-08
Solution

Yaroslav-Stetsko:
image
image
image

### DIFF
--- a/taxi/tests/test_models.py
+++ b/taxi/tests/test_models.py
@@ -1,0 +1,51 @@
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+
+from taxi.models import Manufacturer, Car
+
+
+class ModelsTest(TestCase):
+    def test_manufacturer_str(self):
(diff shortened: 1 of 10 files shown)


HStakhiv (approved):
Looks good to me

Nattalli (changes requested):
Please, update your version with the master's one. Add tests for the searching functionality. Add tests with `post-requests`

taxi/tests/test_models.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_manufacturer_str(self):
+        manufacturer = Manufacturer.objects.create(name="Test", country="test")
+        self.assertEqual(
+            str(manufacturer),
+            f"{manufacturer.name} {manufacturer.country}")
+
+    def test_car_str(self):
+        name = "Test"
+        country = "Test"
+        model = "Test"
```
Nattalli, line 17: why did you decide to create these variables? Also, you have the same value for all of them, so you can create only `test_data="Test"` and use it everywhere

Nattalli (approved):

PR #29, opened 2022-11-08
Solution

luxuriant777:
Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4
Screenshot_5
Screenshot_6


taxi/templatetags/query_transform.py
```diff
@@ -0,0 +1,16 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def query_transform(request, **kwargs):
+    print("request.GET", request.GET)
+    print("kwargs", kwargs)
```
Cerne13, line 9: I guess we won't be needing prints in prod

luxuriant777: Thanks, changed.

TPolina (changes requested):
I don't see tests for search forms and not all views are tested

luxuriant777:
> I don't see tests for search forms and not all views are tested

Changed

TPolina (approved):

PR #32, opened 2022-11-09
Solution

VadymShkarbul:
image
image
image

### DIFF
--- a/taxi/tests/test_admin.py
+++ b/taxi/tests/test_admin.py
@@ -1,0 +1,46 @@
+from django.contrib.auth import get_user_model
+from django.test import TestCase, Client
+from django.urls import reverse
+
+from taxi.models import Car, Manufacturer
+
+
+class AdminSiteTest(TestCase):
+    def setUp(self) -> None:
+        self.client = Client()
+        self.admin_user = get_user_model().objects.create_superuser(
+            username="admin",
+            password="1a2S3r4s5t"
+        )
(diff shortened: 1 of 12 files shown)


taxi/tests/test_admin.py
```diff
@@ -0,0 +1,46 @@
...
+            license_number="RRR54321"
+        )
+        self.manufacturer = Manufacturer.objects.create(
+            name="Ford",
+            country="USA"
+        )
+        self.car = Car.objects.create(
+            model="Focus",
+            manufacturer=self.manufacturer,
+        )
```
HStakhiv, line 28: I think these lines can be removed. What do you think?

taxi/views.py
```diff
@@ -59,6 +84,25 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.all().select_related("manufacturer")
```
HStakhiv, line 85: Delete `all()`

taxi/tests/test_admin.py
VadymShkarbul: Thank you for the review. Yes you are right.

HStakhiv (approved):
Looks good

Nattalli (changes requested):
Please, update your version with the master's one. Add tests for the searching functionality.

Nattalli (approved):
Good job!

PR #42, opened 2022-11-09
Solution

OvDima1:
image
image
image


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,36 @@
...
+                    "password1": "test12345",
+                    "password2": "test12345",
+                    "first_name": "firstname",
+                    "last_name": "lastname",
+                    "license_number": license_number,
+                }
+                form = DriverCreationForm(form_data)
+
+                self.assertFalse(form.is_valid())
+                self.assertNotEqual(form.cleaned_data, form_data)
```
StasOkhrym, line 36: Add tests for search forms

StasOkhrym (approved):

taxi/forms.py
```diff
@@ -3,50 +3,79 @@
...
-class DriverLicenseUpdateForm(forms.ModelForm):
-    class Meta:
-        model = Driver
-        fields = ["license_number"]
+class CarSearchForm(forms.Form):
+    model = forms.CharField(
+        max_length=255,
+        required=False,
+        label="",
+        widget=forms.TextInput(attrs={"placeholder": "Search by model.."})
```
Nattalli, line 62: why do you use 2 dots? I think that 1 or 3 would be better. It is not a mistake, but it can be changed 😅

taxi/views.py
```diff
@@ -59,6 +85,26 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.all().select_related("manufacturer")
```
Nattalli, line 86: remove unnecessary `all()` here and in all queries where it possible

templates/taxi/manufacturer_confirm_delete.html
```diff
@@ -7,4 +7,4 @@ <h1>Delete manufacturer?</h1>
 
     <input type="submit" value="Yes" class="btn btn-danger">
   </form>
-{% endblock %}
+{% endblock %}
```
Nattalli, line 10: ohh, add blank line in the end of the file!

templates/taxi/manufacturer_form.html
```diff
@@ -9,4 +9,4 @@ <h1>{{ object|yesno:"Update,Create" }} manufacturer</h1>
 
     <input type="submit" value="Submit" class="btn btn-primary">
   </form>
-{% endblock %}
+{% endblock %}
```
Nattalli, line 12: and here

Nattalli (changes requested):
you made a good job. Left some comments. Also I have a question: why do you add tests in the folder `tests` and in the file `tests.py`? you can  replace all tests to that folder

Nattalli (approved):

PR #58, opened 2022-11-10
Solution!

ketstap162:


linaSam (changes requested):
Looks nice but you should attach screenshots :)

linaSam (approved):
Great!

taxi/models.py
```diff
@@ -22,7 +22,9 @@ class Meta:
         verbose_name_plural = "drivers"
 
     def __str__(self):
-        return f"{self.username} ({self.first_name} {self.last_name})"
+        if self.first_name and self.last_name:
+            return f"{self.username} ({self.first_name} {self.last_name})"
+        return f"{self.username}"
```
Nattalli, line 27: btw, if there is no `self.first_name` and `self.last_name`, then the __str__ method will return `self.username `, yes?. So do you really need these changes?

taxi/tests.py
```diff
@@ -1,3 +1,3 @@
 from django.test import TestCase
 
-# Create your tests here.
+# Create your testss here.
```
Nattalli, line 3: this is for what? 😅

taxi/views.py
```diff
@@ -59,6 +84,23 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.all().select_related("manufacturer")
```
Nattalli, line 85: remove `.all()` here and everywhere where it is unnecessary

taxi/tests/test_car_views.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class PrivateCarsTests(TestCase):
+    def setUp(self) -> None:
+        self.driver = get_user_model().objects.create_user(
+            username="testuser", password="test123user"
+        )
+        self.client.force_login(self.driver)
+
+        manuf1 = Manufacturer.objects.create(
```
Nattalli, line 24: it is a bad variable name

taxi/tests/test_car_views.py
```diff
@@ -0,0 +1,53 @@
...
+            name="Manufname1",
```
Nattalli, line 25: the same here

Nattalli: that is not difficult to write the full word, isn't it?

taxi/tests/test_driver_views.py
```diff
@@ -0,0 +1,49 @@
...
+        response = self.client.get(DRIVERS_URL)
+
+        for value in [
+            self.driver.id,
+            self.driver.username,
+            self.driver.first_name,
+            self.driver.last_name,
+            self.driver.license_number,
+        ]:
+            self.assertContains(response, value)
```
Nattalli, line 49: can you do the same with testing car views?

taxi/tests/test_models.py
```diff
@@ -0,0 +1,54 @@
...
+from django.test import TestCase
+from django.urls import reverse
+
+from taxi.models import Manufacturer, Driver, Car
+
+
+class ModelsTests(TestCase):
+    def test_manufacturer_str(self):
+        manufacturer = Manufacturer.objects.create(
+            name="Lol kek 221", country="Super great country"
```
Nattalli, line 11: 🥲

Nattalli (changes requested):
Can you also add some tests for `POST requests`? for example, endpoint that creates new car. Also add test the checks if the invalid data wouldn't update the license_number. And the last - add test that checks your searching functionality

taxi/models.py
ketstap162: Yes, because otherwise it will return `username ( )` or `username (first_name )` (with space)

ketstap162: Yes, because otherwise the function will return `username ( )` or `username (first_name )` (with space). This is bad

taxi/tests/test_models.py
ketstap162: ?

Y-Havryliv (approved):
GJ!

PR #60, opened 2022-11-10
Solution

AVyha:
image
image
image
image
image
image
image


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, {"license_number": "QWE12345"})
+
+    def test_validate_license_number(self):
+        values = {
+            "ASD123456": "License number should consist of 8 characters",
+            "aSD12345": "First 3 characters should be uppercase letters",
+            "BKJ1E345": "Last 5 characters should be digits"
+        }
+        for license_number, error in values.items():
```
mltdll, line 49: I think it's better to use subtests here.

taxi/tests/test_views.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class PublicTest(TestCase):
+    def test_login_required_for_main_pages(self):
+        self.assertNotEqual(self.client.get(reverse("taxi:car-list")), 200)
+        self.assertNotEqual(
+            self.client.get(reverse("taxi:manufacturer-list")), 200
+        )
+        self.assertNotEqual(self.client.get(reverse("taxi:driver-list")), 200)
+        self.assertNotEqual(self.client.get(reverse("taxi:index")), 200)
```
mltdll, line 24: Don't test several pages like that. Use subtests or write a separate test for each view.
Chaining assertions results into only one meaningful test result (as opposed to meaningful results or every test) when something fails.

Also, I think it would be justified to use `.assertRedirects()` here.

taxi/tests/test_views.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(
+            list(res.context["manufacturer_list"]), list(manufacturer_list)
+        )
+        self.assertTemplateUsed(res, "taxi/manufacturer_list.html")
+
+    def test_manufacturer_list_page_have_pagination(self):
+        res = self.client.get(reverse("taxi:manufacturer-list"))
+        self.assertEqual(
+            len(res.context["manufacturer_list"]),
+            ManufacturerListView.paginate_by,
```
mltdll, line 48: I guess it's a test of Django built-in functionality. If you specified exact pagination value here, it could considered a test of the exact value — but here you just get the same value you test.

mltdll (changes requested):

mltdll (changes requested):
By the way, don't forget to sync your PR with mate-academy/master and implement tests for search functionality. :)

taxi/tests/test_forms.py
Nattalli: yes. maybe `@pytest.mark.parametrize`?

taxi/tests/test_views.py
Nattalli: yes, @AVyha just make it in different tests

taxi/tests/test_views.py
```diff
@@ -0,0 +1,84 @@
...
+        for i in range(3):
+            Driver.objects.create_user(
+                username=f"username1{i}", password="password12"
+            )
+
+        driver_list = Driver.objects.all()
+        res = self.client.get(reverse("taxi:driver-list"))
+
+        self.assertEqual(list(res.context["driver_list"]), list(driver_list))
+        self.assertTemplateUsed(res, "taxi/driver_list.html")
```
Nattalli, line 84: also, add some tests for post-requests. Something like creating of car, manufacturer by endpoint

Nattalli: also, Write tests for the searching feature that you have implemented. Make sure yourself, that it works as expected.

Nattalli (changes requested):

taxi/tests/test_forms.py
Nattalli: oh no, you use here unittest. You can use `with self.subTest`

PR #61, opened 2022-11-10
"Solution"

rostyak:
driver_list
driver_list_find
car_list
car_list_find
manufacturer_list
manufacturer_list_find

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -17,6 +17,28 @@ class CarForm(forms.ModelForm):
         fields = "__all__"
 
 
+class CarSearchForm(forms.Form):
+    model = forms.CharField(
+        max_length=255,
+        required=False,
+        label="",
+        widget=forms.TextInput(attrs={
+            "placeholder": "Search by model.. 🔎"
+        })
+    )
+
+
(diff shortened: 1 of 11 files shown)


StasOkhrym (changes requested):
I guess it is not enough tests.
Anyway need to add tests for search forms

StasOkhrym (approved):

taxi/forms.py
```diff
@@ -17,6 +17,28 @@ class Meta:
         fields = "__all__"
 
 
+class CarSearchForm(forms.Form):
+    model = forms.CharField(
+        max_length=255,
+        required=False,
+        label="",
+        widget=forms.TextInput(attrs={
+            "placeholder": "Search by model.. 🔎"
```
Nattalli, line 26: why do you use 2 dots? Isn't it more logical to use 3 dots?

Nattalli (changes requested):
also add tests for these cases:
1) tests for `POST request`. For example, endpoint for car creation r license updating
2) tests for searching functionality

Y-Havryliv (approved):
GJ!

PR #70, opened 2022-11-10
'Solution'

Nikyxa:
photo_2022-11-10_17-00-46
photo_2022-11-10_17-00-43
photo_2022-11-10_17-00-40


kateryna3k (changes requested):
please add tests for the search form

PR #126, opened 2023-03-12
Solution

ochernyi:
I don't know why, but when I attach screenshots, raise error (we don't support that file type) thus I use correct file type

### DIFF
--- a/taxi/templatetags/query_transform.py
+++ b/taxi/templatetags/query_transform.py
@@ -1,0 +1,15 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def query_transform(request, **kwarks):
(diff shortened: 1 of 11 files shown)


taxi/templatetags/query_transform.py
```diff
@@ -0,0 +1,15 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def query_transform(request, **kwarks):
```
vasylhnatiuk, line 7: kwargs instead of kwarks)

vasylhnatiuk (changes requested):
please add screenshot with error why you cannot add photo your site here or in slack

vasylhnatiuk (approved):
Well done!

PR #134, opened 2023-03-24
Solution

MaybeTI:


SergiiMachulin (approved):
Why is bigger part of your tests in templatetags folder? And other tests are in other place. Maybe better put them all together in one folder.

YuriiKindrat (approved):

PR #136, opened 2023-03-27
Tag problem

BohdanLazaryshyn:


BohdanLazaryshyn:
I wrote a custom tag for the search driver in the driver list and I'm having a problem with the tag. my template doesn't see this tag
image

NEkropo1:
> I wrote a custom tag for the search driver in the driver list and I'm having a problem with the tag. my template doesn't see this tag image

move templatetags folder to your app_name folder, like here:
image
and all should be fine )

NEkropo1 (changes requested):
So, let's wait )

tests/test_form.py
```diff
@@ -0,0 +1,139 @@
...
+            model="car1_test", manufacturer=self.manufacturer
+        )
+        self.car2 = Car.objects.create(
+            model="car2_test", manufacturer=self.manufacturer
+        )
+        self.car3 = Car.objects.create(
+            model="CAR1_test", manufacturer=self.manufacturer
+        )
+
+    def test_search_car_by_model(self):
```
YuriiKindrat, line 28: u are testing querysets filtering here, the same is for other tests
it is better to test ur search features through views

YuriiKindrat (changes requested):

tests/test_view.py
```diff
@@ -0,0 +1,178 @@
...
+        form_data = {"model": "car1"}
+        form = CarSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        queryset = Car.objects.filter(
+            model__icontains=form.cleaned_data["model"]
+        )
+
+        self.assertIn(self.car1, queryset)
+        self.assertNotIn(self.car2, queryset)
+        self.assertIn(self.car3, queryset)
```
YuriiKindrat, line 34: what do u test here, u are getting filtered queryset, and then just check if car exist in filtered queryset - does this check ur search feature ?
the same is for other tests

YuriiKindrat (approved):

PR #137, opened 2023-03-27
Solution

SergiiMachulin:
Screenshot from 2023-03-27 16-03-27
Screenshot from 2023-03-27 16-03-33
Screenshot from 2023-03-27 16-03-37


WDemidenko (approved):
Good job!

PR #139, opened 2023-03-27
i love django <3

DavidGlusch:

### DIFF
--- a/templates/taxi/car_list.html
+++ b/templates/taxi/car_list.html
@@ -1,12 +1,20 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
+
     Car list
     <a href="{% url 'taxi:car-create' %}" class="btn btn-primary link-to-page">
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
+
+  {{ search_form|crispy }}
+    <input type="submit" class="btn btn-secondary" value="Search">
+
+  </form>
   {% if car_list %}
     <ul>
       {% for car in car_list %}
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -40,7 +40,7 @@ class DriverLicenseUpdateForm(forms.ModelForm):
 
 
 def validate_license_number(
-    license_number,
+        license_number,
 ):  # regex validation is also possible here
     if len(license_number) != 8:
         raise ValidationError("License number should consist of 8 characters")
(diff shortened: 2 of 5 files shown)


yura-hudzovskyi (changes requested):
Where are the tests? What happens if the search returns a lot of pages?

image

bihunva (approved):
Ok

templates/taxi/car_list.html
```diff
@@ -1,12 +1,20 @@
...
     <a href="{% url 'taxi:car-create' %}" class="btn btn-primary link-to-page">
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
+
+  {{ search_form|crispy }}
+    <input type="submit" class="btn btn-secondary" value="Search">
+
+  </form>
```
YuriiKindrat, line 17: will ur search form work correctly with pagination ?

tests/test_forms.py
```diff
@@ -0,0 +1,18 @@
...
+            "username": "test1",
+            "password1": "test234567",
+            "password2": "test234567",
+            "license_number": "AAA12345",
+            "first_name": "test3",
+            "last_name": "test4",
+        }
+        form = DriverCreationForm(data=data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, data)
```
YuriiKindrat, line 18: did u checked here license validation ?

YuriiKindrat (changes requested):

tests/test_views.py
```diff
@@ -0,0 +1,143 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            list(response.context["manufacturer_list"]),
+            list(manufacturers)
+        )
+        self.assertTemplateUsed(response, "taxi/manufacturer_list.html")
+
+    def test_manufacturer_search(self):
+        Manufacturer.objects.create(name="BMW", country="test2")
+        filtered_manufacturer = Manufacturer.objects.filter(
```
YuriiKindrat, line 43: do u use it somewhere ?
also this test doesn't check searched page context 
u are only checking status code + template using, check manufacturers filtered list

YuriiKindrat: the same for other tests
also u can create manufacturers for all tests using setUp

YuriiKindrat (changes requested):

wQuelS (approved):

PR #160, opened 2023-03-28
Solution

VladyslavCherkashyn:
image
image
image
image

### DIFF
--- a/taxi/tests/tests_view.py
+++ b/taxi/tests/tests_view.py
@@ -1,0 +1,67 @@
+from django.contrib.auth import get_user_model
+from django.test import TestCase, Client
+from django.urls import reverse
+
+from taxi.models import Manufacturer
+
+MANUFACTURER_URL = reverse("taxi:manufacturer-list")
+
+
+class PublicManufacturerTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_login_required(self):
+        res = self.client.get(MANUFACTURER_URL)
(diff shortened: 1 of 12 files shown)


AlessaQ (approved):

YuriiKindrat (changes requested):

taxi/tests/tests_view.py
```diff
@@ -0,0 +1,108 @@
...
+        self.client.force_login(self.driver)
+        response = self.client.post(reverse(
+            "taxi:toggle-car-assign",
+            args=[self.car.id])
+        )
+        self.assertEqual(response.status_code, 302)
+        self.assertTrue(self.car in self.driver.cars.all())
+
+        response = self.client.post(reverse(
+            "taxi:toggle-car-assign",
```
wQuelS, line 104: Why do you have only `MANUFACTURER_URL` constant ?

wQuelS (changes requested):

taxi/tests/tests_view.py
VladyslavCherkashyn: Because I used reverse("taxi:manufacturer-list")  more than once or twice for method GET
At the same time, in the case reverse("taxi:toggle-car-assign") I had to add an extra argument to the link which didn`t allow me to move this link into a separate constant (I tried several options, all unsuccessful).
So I have only one constant here - MANUFACTURER_URL.

YuriiKindrat (approved):

PR #161, opened 2023-03-28
Develop

NazarHladaniuk:


Anatolii-Poznyak (approved):

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,18 @@
...
+            "username": "test1",
+            "password1": "test123456",
+            "password2": "test123456",
+            "license_number": "TES12345",
+            "first_name": "first_name",
+            "last_name": "last_name",
+        }
+        form = DriverCreationForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
```
YuriiKindrat, line 18: did u checked here license validation ?

taxi/tests/test_views.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class PrivateManufacturerTest(TestCase):
+    def setUp(self) -> None:
+        self.user = get_user_model().objects.create_user(
+            username="Testusername", password="testpass123456"
+        )
+        self.client.force_login(self.user)
+
+    def test_private_manufacturer(self):
```
YuriiKindrat, line 25: no need to repeat class name here, use retrieve word instead

taxi/tests/test_views.py
```diff
@@ -0,0 +1,138 @@
...
+            list(response.context["manufacturer_list"]),
+            list(manufacturer)
+        )
+        self.assertTemplateUsed(response.context)
+
+    def test_manufacturer_search(self):
+        Manufacturer.objects.create(name="Ford", country="United States")
+        response = self.client.get(MANUFACTURER_URL + "?name=ford")
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "taxi/manufacturer_list.html")
```
YuriiKindrat, line 44: did u test search feature here?
same for others

YuriiKindrat (changes requested):

taxi/tests/test_views.py
NazarHladaniuk: do you mean that it is better to do them in test_forms or what?

YuriiKindrat: i mean u are checking statuscode + template used, are u checked the context of response ? If search feature doesn't work and page will return all cars from db, will this test check this ? can u check search feature by creating only one car for this test ? use setUp for obj creating for all tests 
same for other tests
also add toggle assing me to car test + admin page tests

YuriiKindrat (approved):

PR #168, opened 2023-03-28
solution

AlessaQ:


AlessaQ:
image
image
image
image

tests/test_models.py
```diff
@@ -0,0 +1,48 @@
...
+        self.driver = get_user_model().objects.create_user(
+            username="Test",
+            password="test1234",
+            first_name="Te",
+            last_name="St",
+            license_number="ABC12345"
+        )
+        self.car = Car.objects.create(
+            model="Test",
+            manufacturer=Manufacturer.objects.get(pk=1)
```
YuriiKindrat, line 23: why don't u use self.manufacturer here?

tests/test_models.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+    def test_manufacturer_str(self):
+        manufacturer = self.manufacturer
```
YuriiKindrat, line 27: no need to create variable for attribute, fix everywhere

YuriiKindrat (changes requested):

tests/test_models.py
AlessaQ: fixed everywhere

PR #169, opened 2023-03-28
Solution

oksanaaam:
image
image
image
image

### DIFF
--- a/taxi/tests/test_forms.py
+++ b/taxi/tests/test_forms.py
@@ -1,0 +1,18 @@
+from django.test import TestCase
+
+from taxi.forms import DriverCreationForm, DriverSearchForm
+
+
+class FormsTests(TestCase):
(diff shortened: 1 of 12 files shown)


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,18 @@
...
+            "username": "Marichka",
+            "password1": "z54bvFG543",
+            "password2": "z54bvFG543",
+            "first_name": "Maria",
+            "last_name": "Clinton",
+            "license_number": "ABC12345"
+        }
+        form = DriverCreationForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
```
YuriiKindrat, line 18: what if form doesn't check license number as u want ? will this test check this ?

YuriiKindrat (changes requested):
what about testing search feature ?

PR #170, opened 2023-03-29
Solution

ZhAlexR:
image
image
image
image
image


taxi/tests/test_views.py
```diff
@@ -68,3 +67,25 @@ def test_queryset_with_search_criteria(self):
...
+            model="Golf", manufacturer=self.manufacturer
+        )
+
+    def test_toggle_assign_to_car_driver_in_car_drivers(self):
+        self.client.get(reverse("taxi:toggle-car-assign", args=([1])))
+        self.assertIn(self.user, self.car.drivers.all())
+
+    def test_toggle_assign_to_car_driver_not_in_car_drivers(self):
+        self.client.get(reverse("taxi:toggle-car-assign", args=([1])))
+        self.assertIn(self.user, self.car.drivers.all())
```
YuriiKindrat, line 91: test for checking driver not in car, but u are using assertIn ? both tests are similar

YuriiKindrat (changes requested):

taxi/tests/test_views.py
ZhAlexR: > test for checking driver not in car, but u are using assertIn ? both tests are similar

Yep, my mistake, Thank's

Y-Havryliv (approved):
Nice!

PR #172, opened 2023-03-29
'Solution'

ykopatko:


taxi/tests/test_views.py
```diff
@@ -0,0 +1,135 @@
...
+
+        manufacturers = Manufacturer.objects.filter(name__icontains="F")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            list(response.context["manufacturer_list"]),
+            list(manufacturers)
+        )
+
+        self.assertTemplateUsed(response, "taxi/manufacturer_list.html")
```
YuriiKindrat, line 69: do u create some manufacturers for this tests? all method will also return blank list here)

YuriiKindrat: fix everywhere

YuriiKindrat (changes requested):

taxi/tests/test_views.py
ykopatko: Sorry for being late! It is fixed now.

PR #174, opened 2023-03-29
Develop

Yevheniy-Lototskiy:


taxi/test/test_form.py
```diff
@@ -0,0 +1,61 @@
...
+        Car.objects.create(
+            model="test",
+            manufacturer=manufacturer
+        )
+        cars = Car.objects.all()
+        response = self.client.get(reverse("taxi:car-list") + "?model=test")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "taxi/car_list.html")
+        self.assertEqual(list(response.context["car_list"]), list(cars))
```
YuriiKindrat, line 32: what if search feature won't work ? will this test check this ? 
same for other tests

taxi/test/test_models.py
```diff
@@ -0,0 +1,59 @@
...
+            "password1": "user12345",
+            "password2": "user12345",
+            "first_name": "",
+            "last_name": "",
+        }
+
+        form = DriverCreationForm(data)
+
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, data)
```
YuriiKindrat, line 31: do u create driver here?)

taxi/test/test_models.py
```diff
@@ -0,0 +1,59 @@
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+from django.urls import reverse
+
+from taxi.forms import DriverCreationForm
+from taxi.models import Manufacturer, Car
+
+
+class TestCreateAndDeleteModels(TestCase):
```
YuriiKindrat, line 9: no need to test build in methods here,  u can check some craft methods like str etc.

taxi/test/test_views.py
```diff
@@ -0,0 +1,116 @@
...
+class CarTest(TestCase):
+    def setUp(self):
+        self.user = get_user_model().objects.create_user(
+            username="tset",
+            password="test12345",
+        )
+
+        self.client.force_login(self.user)
+
+    def test_form_should_return_cars(self):
```
YuriiKindrat, line 17: what form should return cars? 
rename ur test, fix everywhere

YuriiKindrat (changes requested):

taxi/test/test_form.py
YuriiKindrat: u didn't fix that, add more cars with different model names

YuriiKindrat (changes requested):

taxi/test/test_form.py
```diff
@@ -0,0 +1,68 @@
...
+        Car.objects.create(
+            model="test3",
+            manufacturer=manufacturer
+        )
+        cars = Car.objects.all()
+        response = self.client.get(reverse("taxi:car-list") + "?model=test3")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "taxi/car_list.html")
+        self.assertEqual(list(response.context["car_list"]), list(cars))
```
YuriiKindrat, line 39: u are testing ur search feature, so u a filtered by model to search test3 model, and then u are compare filtered list with list consisting of all cars am i right ? Does this test ur search feature ? 
fix for all tests

YuriiKindrat (changes requested):
add tests for admin and toggle assign me to car feature

PR #175, opened 2023-03-29
Solution

ghost:


oksana0410:
Taxi Service 2023-03-29 12-41-14

Taxi Service 2023-03-29 12-41-59

Taxi Service 2023-03-29 12-41-37

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+    def test_manufacturer_search(self):
+        search_request = "test_search"
+        response = self.client.get(f"/manufacturers/?name={search_request}")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertQuerysetEqual(
+            response.context["manufacturer_list"],
+            Manufacturer.objects.filter(name__icontains=search_request)
```
YuriiKindrat, line 58: does ur test work normal without any manufacturer/car created ? all method also will reaturn balnk list here)

YuriiKindrat: also why do u need separate variable for serach_request ?

YuriiKindrat (changes requested):
add tests for admin page + license validation form

taxi/tests/test_forms.py
oksana0410: > also why do u need separate variable for serach_request ?

not sure if i understand the question, if yes it is because i use different placeholders

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,39 @@
...
+class SearchFormTest(TestCase):
+    def setUp(self) -> None:
+        self.user = get_user_model().objects.create_user(
+            username="test_username",
+            password="test_password"
+        )
+        self.client.force_login(self.user)
+
+    def test_driver_search(self):
+        search_request = "test_search"
```
YuriiKindrat, line 32: i mean why do u create this var here? why can't u write it directly to the link ? 
also what about search features for other pages ?

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,39 @@
...
+            "username": "test_username",
+            "password1": "test_password",
+            "password2": "test_password",
+            "license_number": "LLL98764",
+            "first_name": "test_first_name",
+            "last_name": "test_last_name"
+        }
+        form = DriverCreationForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
```
YuriiKindrat, line 20: what about test for driver license update form? also it is better to add some invalid data for tests, to check it validation )

YuriiKindrat (changes requested):

Serhii-Leonenko (changes requested):
Add test for `toggle_assign_to_car` view

YuriiKindrat (approved):

PR #176, opened 2023-03-29
'Solution'

NickSavchyn:


tests/test_forms.py
```diff
@@ -0,0 +1,18 @@
...
+            "username": "new_user",
+            "password1": "user123test",
+            "password2": "user123test",
+            "first_name": "test name",
+            "last_name": "test last",
+            "license_number": "ABC12345"
+        }
+        form = DriverCreationForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
```
YuriiKindrat, line 18: add invalid data for this test, will this test check if form doesn't work ?

YuriiKindrat (changes requested):
what about testing ur search feature + toggle assign me to car view?
also why don't u write tests for other views ?
also what about admin page tests ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #177, opened 2023-03-29
solution

usrofgh:


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
+    {{ search_form|crispy }}
+    <input type="submit" value="🔍" class="btn btn-secondary">
+  </form>
```
YuriiKindrat, line 14: will ur search feature work with pagination correctly ?

YuriiKindrat (changes requested):
what about tests?)

Serhii-Leonenko (changes requested):

PR #178, opened 2023-03-29
Solution

Zenionoi:


YuriiKindrat (changes requested):
what about testing ur search feature + toggle assing to car view ?

PR #184, opened 2023-03-31
'Solution'

Vitalii2906:


taxi/views.py
```diff
@@ -36,6 +41,22 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        context["search_form"] = ManufacturerSearchForm(initial={"name": name})
+        return context
+
+    def get_queryset(self):
+        name = self.request.GET.get("name")
+
+        if name:
+            return self.model.objects.filter(name__icontains=name)
+
+        return self.model.objects.all()
```
ArsenPidhoretskyi, line 58:
```suggestion
    def get_queryset(self):
        queryset = self.queryset
        name = self.request.GET.get("name")

        if name:
            queryset = queryset.filter(name__icontains=name)

        return queryset
```

ArsenPidhoretskyi: Can you check if this works?

ArsenPidhoretskyi: The same in other places

templates/taxi/manufacturer_list.html
```diff
@@ -1,8 +1,13 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
     Manufacturer List
+    <form action="" method="get">
+      {{ search_form|crispy }}
+      <input type="submit">
+    </form>
```
ArsenPidhoretskyi, line 10: Duplicate of code. Please create new include html file for this

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #201, opened 2023-04-28
Solution

YehorKorn:
image
image
image
image

### DIFF
--- a/templates/taxi/driver_list.html
+++ b/templates/taxi/driver_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
     <h1>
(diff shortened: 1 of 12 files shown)


templates/taxi/driver_list.html
```diff
@@ -8,6 +9,12 @@ <h1>
       </a>
     </h1>
 
+    <form action="" method="get" class="form-inline">
+      {{ search_form|crispy }}
+
+      <input style="margin-left: 10px" type="submit" value="🔎" class="btn btn-outline-primary">
+    </form>
+
```
ArsenPidhoretskyi, line 17: No need to duplicate code. Better to create additional template and than include it

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #210, opened 2023-05-21
Solution

aarrtemm:
Search driver:
image

Search car:
image

Search Manufacturer:
image

### DIFF
--- a/templates/taxi/manufacturer_list.html
+++ b/templates/taxi/manufacturer_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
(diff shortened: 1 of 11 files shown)


templates/taxi/manufacturer_list.html
```diff
@@ -7,7 +8,9 @@ <h1>
       Create
     </a>
   </h1>
-
+  <form action="" method="get">
+    {{ search_manufacturer_form|crispy }}
+  </form>
```
Vasyl-Smutok, line 13: Move this form into another template(includes) and then reuse it in all places.

Vasyl-Smutok (changes requested):

vasylhnatiuk (approved):

PR #264, opened 2023-07-18
solution

OksanaAdamchuk:

### DIFF
--- a/templates/taxi/car_list.html
+++ b/templates/taxi/car_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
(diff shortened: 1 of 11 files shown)


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
+    {{ search_form|crispy }}
+    <input type="submit" value="search" class="btn btn-primary">
+  </form>
```
ArsenPidhoretskyi, line 14: Move this to separate template and reuse it by include tag

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Good Job 👍

PR #278, opened 2023-07-20
Solution

Po-SvinSkipodla:
image
image
image


Po-SvinSkipodla:
I video lessons, Danya said that if we don't use admin page often, we can skip admin tests, but if in this task it needs to be done, i will

taxi/tests/test_views.py
Oleg-s-s, line 1: What about creating/updating and other views?

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Sorry, its ok

PR #296, opened 2023-08-08
'Solution'

timyr4iq:
image
image
image

### DIFF
--- a/templates/taxi/car_list.html
+++ b/templates/taxi/car_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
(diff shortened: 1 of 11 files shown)


templates/taxi/car_list.html
```diff
@@ -7,6 +8,12 @@ <h1>
       Create
     </a>
   </h1>
+
+  <form action="" method="get" class="form-inline">
+    {{ search_form|crispy }}
+    <input type="submit" value="🔎" class="btn btn-secondary">
+  </form>
```
ArsenPidhoretskyi, line 15: Move this tp separate html and just include in all needed templates

ArsenPidhoretskyi (changes requested):

PR #316, opened 2023-09-27
Solution

CaCtUs1k:
image
image
image


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
```
innyshka, line 11: maybe it will be better if you put this form in a separate file in "includes". Because you use the same form three times

innyshka (approved):

eduardhabryd (approved):
LGTM!

Sofbog (approved):
LGTM

PR #343, opened 2023-10-14
Solution

ochernous:


obohatov (approved):
NJ!

taxi/views.py
```diff
@@ -36,6 +43,22 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["name"] = name
```
Sofbog, line 49: what purpose of this line?

Sofbog (changes requested):

taxi/views.py
ochernous: I thought we should reassign it

PR #351, opened 2023-10-28
Solution

ilizh:
Снимок экрана 2023-10-28 160906
Снимок экрана 2023-10-28 160900
Снимок экрана 2023-10-28 160854

### DIFF
--- a/templates/taxi/manufacturer_list.html
+++ b/templates/taxi/manufacturer_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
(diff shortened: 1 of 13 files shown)


ArsenPidhoretskyi (changes requested):
Move this part to separate form and reuse it

templates/taxi/manufacturer_list.html
```diff
@@ -7,7 +8,10 @@ <h1>
       Create
     </a>
   </h1>
-
+	<form method="get" action="">
```
ArsenPidhoretskyi, line 11: Here

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
good job! But in future think about: do you need to have search form on every page, which extends base?

PR #366, opened 2023-11-21
Solution

Daniil-Pankieiev:


OlehShumov (approved):

taxi/tests/test_admin.py
```diff
@@ -0,0 +1,36 @@
...
+            username="testuser",
+            password="testpassword",
+            first_name="Test_first",
+            last_name="test_last",
+            license_number="123456",
+        )
+
+    def test_driver_display(self):
+        url = reverse("admin:taxi_driver_changelist")
+        res = self.client.get(url)
```
LiudmylaKulinchenko, line 23: use descriptive names, change res to result or smth

taxi/tests/test_form.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertFalse(form2.is_valid())
+
+    def test_driver_wrong_len_license(self):
+        form_data = {
+            "username": "test1user",
+            "password1": "test1pass",
+            "password2": "test1pass",
+            "license_number": "ABC12345678",
+            "first_name": "first1",
+            "last_name": "last1",
```
LiudmylaKulinchenko, line 87: try to avoid code duplicating

LiudmylaKulinchenko (changes requested):

taxi/tests/test_form.py
Daniil-Pankieiev: Is it okay to use this way?

LiudmylaKulinchenko (approved):
Great 🔥

taxi/tests/test_form.py
LiudmylaKulinchenko: Yes, looks good

PR #385, opened 2023-11-21
sol

oleksandr-pupchenko:


fedorov-s-od (approved):

taxi/tests.py
```diff
@@ -1,3 +1,108 @@
...
+        ("taxi:manufacturer-update", 1),
+        ("taxi:manufacturer-delete", 1),
+        ("taxi:car-update", 1),
+        ("taxi:car-delete", 1),
+        ("taxi:driver-update", 1),
+        ("taxi:driver-delete", 1)
+    ]
+
+    @parameterized.expand(list_and_create_urls)
+    def test_authenticated_user(self, view_name):
```
LiudmylaKulinchenko, line 79: Test name is not clear. You should name it according to a feature you want to test. Looks now like you want to test if user can be authenticated

LiudmylaKulinchenko (changes requested):

PR #395, opened 2023-11-29
'Solution'

ArtemPutsenko:


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit" value="🔍">
+  </form>
```
EdAlekseiev, line 14: Create separate template for search form and use it where you need

EdAlekseiev (changes requested):

templates/taxi/car_list.html
ArtemPutsenko: How is that? I did not quite understand

EdAlekseiev: You have separate template for pagination and sidebar. You need to create the same template for search form and use it in other templates

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
Also, you must write tests for your code according to task.

EdAlekseiev (approved):
Good job

PR #422, opened 2024-01-16
Solution

kravetb:
Task8(1)
Task8(2)
Task8(3)


kravetb:
How many test I need write there and which test?

NikOneZ1 (approved):

PR #457, opened 2024-02-28
Solution

kabust:
Drivers search:

image

Cars search:

image

Manufacturers search:

image


taxi/tests/test_views.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+        self.client.force_login(self.user)
+
+    def test_manufacturer_list_correct_status(self):
+        response = self.client.get(reverse("taxi:manufacturer-list"))
+        self.assertEqual(response.status_code, 200)
+
+    def test_manufacturer_list_pagination(self):
+        response = self.client.get(reverse("taxi:manufacturer-list"))
+        self.assertEqual(response.context["manufacturer_list"].count(), 5)
```
aLEKS-e3, line 73: do you really need to test this? I guess since it's a built-in option, everything should be fine

taxi/tests/test_models.py
```diff
@@ -0,0 +1,50 @@
+from django.test import TestCase
+
+from taxi.models import Manufacturer, Car, Driver
+
+
+class ManufacturerModelTest(TestCase):
+    def test_manufacturer(self):
```
aLEKS-e3, line 7: I think it's better to mention that you test str method here)

aLEKS-e3 (changes requested):

aLEKS-e3 (approved):
seems goood!

taxi/views.py
```diff
@@ -59,6 +83,22 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.all().select_related("manufacturer")
 
+    def get_queryset(self):
+        model = self.request.GET.get("model")
+        queryset = Car.objects.all()
```
NEkropo1, line 88: since you override default queryset with get_queryset, you should remove default queryset since its overriden, also, why you removed select_related manufacturers?

taxi/views.py
```diff
@@ -85,6 +125,24 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
     paginate_by = 5
 
+    def get_queryset(self):
```
NEkropo1, line 128: model and queryset are interchangeable so if you overriden get_queryset, which returns queryset for view, there's no need in model attribute, you can remove it.

NEkropo1 (changes requested):

NEkropo1 (approved):
LGTM!

PR #474, opened 2024-03-05
Solution

AntonSkyrda:
image
image
image


kstorozhenko (approved):

NEkropo1 (changes requested):
remove model field everywhere where you provide get_queryset, since it overrides behaviour of model field and we don't need them more.

Y-Havryliv (approved):

PR #485, opened 2024-03-08
Solution

u123dev:


u123dev:
tf01
tf02
tf03
tf04
tf05
tf06

YukihiroSM (approved):
Everything looks pretty good to me! 
The only comment: tests -> test_form.py should be named as test_forms.py, because you are testing several forms there. But that's not really an issue.

PR #487, opened 2024-03-10
Solution

Serejka21:
image

image

image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,16 +1,22 @@
+from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseRedirect, HttpRequest, HttpResponse
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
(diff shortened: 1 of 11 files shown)


taxi/views.py
```diff
@@ -59,6 +80,17 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.select_related("manufacturer")
```
Serhii-Leonenko: you can delete queryset attr cause you override get_queryset method

taxi/views.py
```diff
@@ -85,6 +117,21 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(DriverListView, self).get_context_data(**kwargs)
```
Serhii-Leonenko: use super without attributes, `context = super().get_context_data(**kwargs)`

taxi/views.py
```diff
@@ -85,6 +117,21 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
```
Serhii-Leonenko: use *args instead of *

Serhii-Leonenko (changes requested):
Add tests for views, check search logic in it, check permissions, check logic custom index and toggle_assign_to_car

Serhii-Leonenko (approved):

PR #499, opened 2024-03-27
Solution

vsupruniuk:


vsupruniuk:
image
image
image

tests/test_views.py
```diff
@@ -0,0 +1,258 @@
...
+
+from taxi.models import Manufacturer, Car
+
+
+class PublicViewsTests(TestCase):
+    def test_index_view_login_required(self) -> None:
+        url = reverse("taxi:index")
+        res = self.client.get(url)
+
+        self.assertNotEqual(res.status_code, 200)
```
LiudmylaKulinchenko: You should check if status code is equal to forbidden instead. Now the test will pass even is server error occurred. Same for the next tests

LiudmylaKulinchenko (changes requested):

tests/test_views.py
vsupruniuk: Probably we should check for 302 status code, cause if user unauthorized, django will redirect user to login page with 302 status, also checking for 401/403 doesn't work

LiudmylaKulinchenko: I agree!

LiudmylaKulinchenko (approved):
Well done 🙌

PR #505, opened 2024-03-31
Solution

Kavbian:
image
image
image

### DIFF
--- a/templates/taxi/car_list.html
+++ b/templates/taxi/car_list.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load crispy_forms_filters %}
 
 {% block content %}
   <h1>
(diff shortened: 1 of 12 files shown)


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit" value="🔎">
+  </form>
```
ArsenPidhoretskyi: Please move this template to separate template and then reuse it in all places(Use include)

ArsenPidhoretskyi (changes requested):
What about search tests?

EdAlekseiev (approved):
Good job

PR #523, opened 2024-05-01
'Solution'

ArtomUmantsev:


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+    <form method="get" action="" class="form-inline">
+      {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit" value="🔎">
+    </form>
```
EdAlekseiev: Create separate template for search form and use where you need to avoid code duplication

EdAlekseiev (changes requested):

vsmutok (approved):
Good Job!

PR #535, opened 2024-05-14
Dev

B1GL3S:


templates/taxi/car_list.html
```diff
@@ -7,6 +7,12 @@ <h1>
       Create
     </a>
   </h1>
+
+  <form action="" method="get" class="form-inline">
+    {{ search_form.as_p }}
+    <input type="submit">
+  </form>
```
ArsenPidhoretskyi: Move this to separate template and reuse it in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #540, opened 2024-05-14
solution: search forms and tests

panicua:
Pages save the search field filter(you can see it's 2nd page and search filter still on), the forms save previous input.

chrome_s8uoFwp4Vp
chrome_Q6BDMHoO78
chrome_lWC1kCurvQ

### DIFF
--- a/taxi/tests/test_views.py
+++ b/taxi/tests/test_views.py
@@ -1,5 +1,5 @@
 from django.contrib.auth import get_user_model
-from django.test import TestCase, Client
+from django.test import TestCase
 from django.urls import reverse
 
 from taxi.models import Manufacturer, Car
@@ -12,7 +12,7 @@ DRIVER_URL = reverse("taxi:driver-list")
 class PublicManufacturerTest(TestCase):
     def test_manufacturer_login_required(self):
         response = self.client.get(MANUFACTURER_URL)
-        self.assertNotEqual(response.status_code, 200)
+        self.assertEqual(response.status_code, 302)
 
 
 class PrivateManufacturerTest(TestCase):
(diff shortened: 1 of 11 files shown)


taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+
+MANUFACTURER_URL = reverse("taxi:manufacturer-list")
+CAR_URL = reverse("taxi:car-list")
+DRIVER_URL = reverse("taxi:driver-list")
+
+
+class PublicManufacturerTest(TestCase):
+    def test_manufacturer_login_required(self):
+        response = self.client.get(MANUFACTURER_URL)
+        self.assertNotEqual(response.status_code, 200)
```
LiudmylaKulinchenko: Better to check if it's equal specific status code in this case. != 200 means there could be any other status code, which is not really covers the case

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+            list(response.context["manufacturer_list"]),
+            list(manufacturers)
+        )
+        self.assertTemplateUsed(response, "taxi/manufacturer_list.html")
+
+
+class PublicCarTest(TestCase):
+    def test_car_login_required(self):
+        response = self.client.get(CAR_URL)
+        self.assertNotEqual(response.status_code, 200)
```
LiudmylaKulinchenko: same here

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+            list(response.context["car_list"]),
+            list(cars)
+        )
+        self.assertTemplateUsed(response, "taxi/car_list.html")
+
+
+class PublicDriverTest(TestCase):
+    def test_driver_login_required(self):
+        response = self.client.get(DRIVER_URL)
+        self.assertNotEqual(response.status_code, 200)
```
LiudmylaKulinchenko: and here

LiudmylaKulinchenko (commented):
Create tests for searching

LiudmylaKulinchenko (changes requested):
forgot to make a reject 👀

panicua:
> Create tests for searching

I forgot about them ;O

YuriiKindrat (approved):
GJ

PR #541, opened 2024-05-14
solution

alonasorochynska:


taxi/tests/test_car_view.py
```diff
@@ -0,0 +1,44 @@
...
+        response = self.client.get(CAR_URL)
+        self.assertEqual(response.status_code, 200)
+        cars = Car.objects.all()
+        self.assertEqual(
+            list(response.context["car_list"]),
+            list(cars),
+        )
+        self.assertTemplateUsed(
+            response, "taxi/car_list.html"
+        )
```
YuriiKindrat: what about search tests ? same is for other views
also it is better to add 'View' to the test naming
PrivateCarViewTest

YuriiKindrat (changes requested):

MNDonut (approved):

PR #552, opened 2024-05-17
Solution

SerhiiMar:
image
image
image

### DIFF
--- a/taxi/tests/test_forms.py
+++ b/taxi/tests/test_forms.py
@@ -1,6 +1,7 @@
 from django.contrib.auth import get_user_model
 from django.forms import CheckboxSelectMultiple
 from django.test import TestCase
+from django.urls import reverse
 
 from taxi.forms import (
     ManufacturerSearchForm,
@@ -10,7 +11,7 @@ from taxi.forms import (
     DriverLicenseUpdateForm,
     DriverSearchForm,
 )
-from taxi.models import Manufacturer
+from taxi.models import Manufacturer, Driver
 
 
 class ManufacturerFormsTest(TestCase):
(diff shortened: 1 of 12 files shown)


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,146 @@
...
+    def test_driver_creation_form(self):
+        form = DriverCreationForm(data=self.creation_form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, self.creation_form_data)
+
+    def test_driver_creation_form_with_invalid_license_number(self):
+        for invalid_license_number in self.invalid_license_numbers:
+            self.creation_form_data["license_number"] = invalid_license_number
+            form = DriverCreationForm(data=self.creation_form_data)
+            self.assertFalse(form.is_valid())
```
LiudmylaKulinchenko: Try to remove loops here and in test_driver_license_update_form_with_invalid_license_number test. One test should be for one case
You can just check if driver creates/updates successfully with license_number and create other tests for each validation case (license_number too short, 3 first should be uppercase letters, etc.)

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+        paths = [
+            "/",
+            "/cars/",
+            "/drivers/",
+            "/manufacturers/",
+        ]
+        for path in paths:
+            response = self.client.get(path)
+            self.assertEqual(response.status_code, 302)
+            self.assertRedirects(response, f"/accounts/login/?next={path}")
```
LiudmylaKulinchenko: same here, rewrite without loop
You also should use reverse instead of hardcoding paths and test redirect only using one path (only car-list, for example)

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #592, opened 2024-07-08
Develop

Timur5050:


pavlopro (changes requested):
Attach screenshots according to task requirements

Timur5050:
image
image
image
image

pavlopro (changes requested):
Just found that you modified migration file
You shouldn't do this

Timur5050:
i modified only for flake8 rules, i have just changed ' for "

pavlopro (approved):
Yes, I understand you
In future those type of files you should add to excluded flake8 files and not modify them

Timur5050:
okay, i got it. so, what do i need to do now?

PR #600, opened 2024-07-14
Solution

BohdanKuzik:

### DIFF
--- a/taxi/tests/test_manufacturer_views.py
+++ b/taxi/tests/test_manufacturer_views.py
@@ -1,0 +1,51 @@
+from django.test import TestCase
+from django.urls import reverse
+from django.contrib.auth import get_user_model
+from taxi.models import Manufacturer
+
+manufacturer_url = reverse("taxi:manufacturer-list")
+
+
+class PublicManufacturerTest(TestCase):
+    def setUp(self):
+        self.url = manufacturer_url
+
+    def test_login_required(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 302)
+        self.assertTrue(response.url.startswith("/accounts/login/"))
+
+
+class PrivateManufacturerListViewTest(TestCase):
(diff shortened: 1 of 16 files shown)


taxi/tests/test_manufacturer_views.py
```diff
@@ -0,0 +1,51 @@
...
+            list(response.context["manufacturer_list"]), list(manufacturers)
+        )
+        self.assertTemplateUsed(response, "taxi/manufacturer_list.html")
+
+    def test_context_contains_correct_data(self):
+        response = self.client.get(self.url)
+        self.assertIsNotNone(response.context["manufacturer_list"])
+        self.assertQuerysetEqual(
+            response.context["manufacturer_list"],
+            Manufacturer.objects.all().order_by("name"),
```
Arsen-hrynevych, line 44: When you use `order_by`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `.all()`.

taxi/tests/test_manufacturer_views.py
```diff
@@ -0,0 +1,51 @@
...
+        self.client.force_login(self.user)
+        self.url = manufacturer_url
+
+        Manufacturer.objects.create(name="Toyota", country="Japan")
+        Manufacturer.objects.create(name="Ford", country="USA")
+
+    def test_retrieve_manufacturers(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        manufacturers = Manufacturer.objects.all().order_by("name")
```
Arsen-hrynevych, line 33: When you use `order_by`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `objects.all()`.

Arsen-hrynevych (approved):
Good job! I really appreciate your effort. Please review the checklist for the task. Also, please read more about the DRY principle.

PR #628, opened 2024-07-31
Implement search & tests

anton-bilous:
image
image
image

### DIFF
--- a/taxi/tests.py
+++ b/taxi/tests.py
@@ -59,16 +59,14 @@ class TestDriverCreationForm(TestCase):
         self.assertFalse(form.is_valid())
 
 
-class TestManufacturerModel(TestCase):
-    def test_str(self):
+class TestModels(TestCase):
+    def test_manufacturer_str(self):
         name = "Test"
         country = "Testland"
         manufacturer = Manufacturer.objects.create(name=name, country=country)
         self.assertEqual(str(manufacturer), f"{name} {country}")
 
-
-class TestCarModel(TestCase):
-    def test_str(self):
+    def test_car_str(self):
         model = "Test"
         manufacturer = Manufacturer.objects.create(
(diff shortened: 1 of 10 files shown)


taxi/tests.py
```diff
@@ -1,3 +1,317 @@
...
+
+class PublicTestCarDeleteView(TestCase):
+    def test_login_required(self):
+        manufacturer = Manufacturer.objects.create(
+            name="Test", country="Testland"
+        )
+        car = Car.objects.create(model="Test", manufacturer=manufacturer)
+        url = reverse("taxi:car-delete", args=(car.id,))
+        response = self.client.get(url)
+        self.assertNotEqual(response.status_code, 200)
```
MNDonut: It's bad to have many test cases with 1 test because each class that inherits from TestCase triggers "migrate" command to create db. Thus if your project has 100 migrations, and 10 test cases, then 100 migrations will be run 10 times and it slows down speed up + deploy. Let's merge them.
Maybe you called the test command with --parallel option, but it won't help here because you have exactly more test cases than your laptop has number of cores

taxi/views.py
```diff
@@ -36,6 +43,24 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        form = ManufacturerSearchForm(self.request.GET)
+
+        queryset = Manufacturer.objects.all()
```
MNDonut: Call super method to get an initial queryset because otherwise it's useless to set model attribute. Fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #643, opened 2024-08-19
Implemented solution

dimak20:


dimak20:
image
image
image
image
image
image
image

taxi/views.py
```diff
@@ -57,7 +79,25 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
     paginate_by = 5
-    queryset = Car.objects.select_related("manufacturer")
+
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(CarListView, self).get_context_data(**kwargs)
```
MNDonut: for get_context_data do you need to specify class in case only one of the parent classes has this method?
fix everywhere

MNDonut (changes requested):

taxi/views.py
dimak20: We did the same in lessons. 
Ok

MNDonut (approved):

PR #647, opened 2024-08-28
search its well!

TsvetanKichuk:


taxi/views.py
```diff
@@ -6,7 +6,8 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm, DriverSearchForm, CarSearchForm, \
+    ManufacturerSearchForm
```
Dimosphen1, line 10: Place all imports on a new line inside of `()`, so there would be one import per line and no imports on the same line with `(` and `)`

Dimosphen1 (changes requested):
Cover your project with tests according to the task description. Make sure that all tests pass.

PR #649, opened 2024-08-31
Develop

TsvetanKichuk:
Снимок экрана 2024-08-31 194614
Снимок экрана 2024-08-31 194622
Снимок экрана 2024-08-31 194637
Снимок экрана 2024-08-31 194735


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,41 @@
+from taxi.forms import DriverLicenseUpdateForm
+
+from django.test import TestCase
```
Dimosphen1, line 3: Order your imports according to PEP-8 in the future

taxi/views.py
```diff
@@ -56,9 +79,25 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
-    paginate_by = 5
+    paginate_by = 1
```
Dimosphen1, line 82: It does not make much sense to reduce the number of elements that are displayed on a page to 1, leave this attribute with a greater value in the future

Dimosphen1 (approved):
LGTM, several comments are left for consideration

taxi/views.py
TsvetanKichuk: i know but i had few elements for pagination in 5 elements on the page

PR #650, opened 2024-09-01
Solution

DSahalatyi:
**Search by username implemented on drivers page**
image

**Search by model implemented on cars page**
image

**Search by name implemented on manufacturers page**
image

**All searches support pagination**
image

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -49,7 +49,7 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
         return context
 
     def get_queryset(self):
-        queryset = Manufacturer.objects.all()
+        queryset = super().get_queryset()
         form = ManufacturerSearchForm(self.request.GET)
         if form.is_valid():
             return queryset.filter(name__icontains=form.cleaned_data["name"])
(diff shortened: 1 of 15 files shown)


taxi/views.py
```diff
@@ -32,10 +39,22 @@ def index(request):
...
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name")
+        context["name"] = name
+        context["search_form"] = ManufacturerSearchForm(initial={"name": name})
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
```
MNDonut: call super method to get an initial queryset. fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -85,6 +118,24 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(DriverListView, self).get_context_data(**kwargs)
+        username = self.request.GET.get("username")
```
MNDonut: Do you need to define class in super when only one parent class has this method? fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #654, opened 2024-09-05
Solution

OleksiukStepan:
Screenshot 2024-09-05 at 15 16 28
Screenshot 2024-09-05 at 15 15 18
Screenshot 2024-09-05 at 15 14 59

### DIFF
--- a/taxi/tests/test_forms.py
+++ b/taxi/tests/test_forms.py
@@ -1,6 +1,6 @@
 from django.contrib.auth import get_user_model
 from django.test import TestCase
-
+from django.urls import reverse
 from taxi.forms import (
     CarForm,
     CarSearchForm,
@@ -9,7 +9,7 @@ from taxi.forms import (
     DriverLicenseUpdateForm,
     ManufacturerSearchForm,
 )
-from taxi.models import Manufacturer
+from taxi.models import Manufacturer, Car
 
 
 class FormTests(TestCase):
@@ -23,6 +23,15 @@ class FormTests(TestCase):
         self.driver2 = get_user_model().objects.create_user(
             username="driver2", password="test321", license_number="DRV12345"
         )
+
+        self.car1 = Car.objects.create(
+            model="Skyline", manufacturer=self.manufacturer
+        )
+        self.car2 = Car.objects.create(
+            model="Civic", manufacturer=self.manufacturer
+        )
+        self.car1.drivers.set([self.driver])
+        self.car2.drivers.set([self.driver2])
 
     def test_car_form_valid_data(self):
         form_data = {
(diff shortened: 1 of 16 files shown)


taxi/views.py
```diff
@@ -36,6 +44,19 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(initial={"name": name})
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
```
MNDonut: call super method to get an initial queryset

taxi/views.py
```diff
@@ -59,6 +80,18 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.select_related("manufacturer")
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(CarListView, self).get_context_data(**kwargs)
```
MNDonut: does LoginRequiredMixin has this method, so that you should specify from which class you should call this method?

MNDonut (changes requested):

taxi/views.py
```diff
@@ -59,6 +80,18 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        model = self.request.GET.get("model", "")
+        context["search_form"] = CarSearchForm(initial={"model": model})
+        return context
+
+    def get_queryset(self):
+        model = self.request.GET.get("model")
+        if model:
+            return Car.objects.filter(model__icontains=model)
```
MNDonut: call super method. fix everywhere

taxi/views.py
```diff
@@ -85,6 +118,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(DriverListView, self).get_context_data(**kwargs)
```
MNDonut: what about this one? fix everywhere

MNDonut (changes requested):

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(form.cleaned_data, form_data)
+
+    def test_valid_license_number_update(self):
+        form_data = {"license_number": "TST12345"}
+        form = DriverLicenseUpdateForm(data=form_data, instance=self.driver)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
+
+    def test_car_search_form(self):
+        form = CarSearchForm(data={"model": "Skyline"})
```
Oleksl888, line 65: Search test functionality must be done on bigger data sets + you need to assert that correct results are present in response and incorrect results are not there. Right now testing if the form is valid is only checking if search query has been passed to the form

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Oleksl888: change relative imports here

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888: this form repeats on 3 different templates, can you make it DRY?

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done! Couple of notes:
1. When using black or similar tools exclude files like migrations.py
2. Is better to use fixtures to load test data - this makes your tests more readable and structures code better

PR #656, opened 2024-09-07
Solution

oleksashcherbakov:


oleksashcherbakov:
image
image
image

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
pavlopro: Define separate template for search and include it into places you need

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #661, opened 2024-09-10
Test

dartomOOv:


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,88 @@
...
+        expected_result = "model=a"
+        data = {"model": "a"}
+        response = self.client.get(reverse("taxi:car-list"), data)
+        self.assertEqual(expected_result, response.request["QUERY_STRING"])
+
+    def test_manufacturer_search_form(self):
+        expected_result = "name=a"
+        data = {"name": "a"}
+        response = self.client.get(reverse("taxi:manufacturer-list"), data)
+        self.assertEqual(expected_result, response.request["QUERY_STRING"])
```
MNDonut: you test if query string is correct, but it doesn't guarantee that the results are filtered correctly

taxi/views.py
```diff
@@ -36,6 +43,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        form = ManufacturerSearchForm(self.request.GET)
+        queryset = Manufacturer.objects.all()
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
+        return queryset
```
fix everywhere

MNDonut (changes requested):

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self.assertEqual(expected_url_result, url_result)
+        self.assertEqual(list(expected_query_result), list(query_result))
+
+    def test_car_search_form(self):
+        Car.objects.create(
+            model="testmodel",
+            manufacturer=self.manufacturer
+        )
+        expected_url_result = "model=o"
```
Oleksl888: Search test functionality must be done on bigger data sets. Right now you are testing on only one record which does not guarantee correct results

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888: search form repeats on 3 templates. Can you optimize it ?

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!
Few notes for the tests:
1. If you compare lists for equality like here `self.assertEqual(list(expected_query_result), list(query_result))` is better to call sorted() on every list to be sure that they are ordered
2. Is better to use fixtures to load test data - makes your tests cleaner and better structured

PR #663, opened 2024-09-10
Developer

AndrZhuk:


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+  {{ search_form|crispy }}
+  <input class="btn btn-secondary" type="submit" value="📞">
+  </form>
```
vsmutok: Create a separate template for search form and include it where you need to avoid code duplication

vsmutok (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #665, opened 2024-09-14
Solution

olena4reunited:


taxi/views.py
```diff
@@ -85,6 +121,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        return context
+
+    def get_queryset(self):
+        queryset = Driver.objects.all()
+        form = DriverSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(
+                username__icontains=form.cleaned_data["username"]
+            )
+        return queryset
```
fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -162,11 +164,11 @@ class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
...
-    ):  # probably could check if car exists
-        driver.cars.remove(pk)
+    driver = get_object_or_404(Driver, id=request.user.id)
+    car = get_object_or_404(Car, id=pk)
+
+    if car in driver.cars.all():
+        driver.cars.remove(car)
     else:
-        driver.cars.add(pk)
+        driver.cars.add(car)
```
MNDonut: use .exists instead of "in" operator because for "in" you should fetch all records and loop through them. instead of it you can do it via SQL

MNDonut (changes requested):

MNDonut (approved):

PR #666, opened 2024-09-16
solution

BastovOleksandr:
car list page with search by model form
car list search by model form works
driver list page with search by username form
driver list search by username form works
manufacturer list page with search by name form
manufacturer list search by name form works

### DIFF
--- a/taxi/tests/test_manufacturer_model.py
+++ b/taxi/tests/test_manufacturer_model.py
@@ -1,0 +1,40 @@
+from django.test import TestCase
+
+from taxi.models import Manufacturer, Car
+from taxi.tests import fixtures
+
+
+class ManufacturerModelTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        fixtures.fill_db()
+
+    def test_object_to_str(self):
+        manufacturer = Manufacturer.objects.get(id=1)
+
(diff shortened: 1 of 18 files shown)


taxi/tests/test_manufacturer_model.py
```diff
@@ -0,0 +1,40 @@
+from django.test import TestCase
+
+from taxi.models import Manufacturer, Car
+from taxi.tests import fixtures
+
+
+class ManufacturerModelTests(TestCase):
+    @classmethod
```
MNDonut, line 8: no need to use this decorator. fix everywhere

taxi/views.py
```diff
@@ -36,6 +44,26 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+    def get_queryset(self):
+        form = ManufacturerNameSearchForm(self.request.GET)
+        self.queryset = Manufacturer.objects.all()
+
+        if form.is_valid():
+            return self.queryset.filter(
+                name__icontains=form.cleaned_data["name"]
+            )
+
+        return self.queryset
```
fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #667, opened 2024-09-16
implement taxi-service-search-and-test

ValeriiaKyr:
Снимок экрана 2024-09-16 в 16 53 42
Снимок экрана 2024-09-16 в 16 54 04
Снимок экрана 2024-09-16 в 16 54 18

### DIFF
--- a/taxi/tests/test_form.py
+++ b/taxi/tests/test_form.py
@@ -11,17 +11,48 @@ from taxi.models import Driver, Car, Manufacturer
 
 class FormsTest(TestCase):
     def setUp(self):
-        self.manufacturer1 = Manufacturer.objects.create(name="test", country="testcountry")
-        self.manufacturer2 = Manufacturer.objects.create(name="another", country="anothercountry")
+        self.manufacturer1 = (
+            Manufacturer.objects.create(
+                name="test",
+                country="testcountry"
+            )
+        )
+        self.manufacturer2 = (
+            Manufacturer.objects.create(
+                name="another",
+                country="anothercountry"
+            )
+        )
 
-        self.car1 = Car.objects.create(model="test", manufacturer=self.manufacturer1)
-        self.car2 = Car.objects.create(model="another", manufacturer=self.manufacturer2)
+        self.car1 = (
+            Car.objects.create(
+                model="test",
+                manufacturer=self.manufacturer1
+            )
+        )
+        self.car2 = (
+            Car.objects.create(
+                model="another",
+                manufacturer=self.manufacturer2
+            )
+        )
 
-        self.driver1 = Driver.objects.create_user(username="testuser", password="pass12345", license_number="ABC12345")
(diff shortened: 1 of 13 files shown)


taxi/views.py
```diff
@@ -36,6 +43,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
+        return queryset
```
MNDonut: 1. call super method to get an initial queryset in get_queryset
2. for get_context_data do you need to specify class in case only one of the parent classes has this method?
fix everywhere

MNDonut (changes requested):

taxi/tests/test_form.py
```diff
@@ -0,0 +1,41 @@
...
+            "password2": "user12test",
+            "first_name": "test first",
+            "last_name": "test last",
+            "license_number": "TES12345",
+        }
+        form = DriverCreationForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, form_data)
+
+    def test_drivers_search_form(self):
```
Oleksl888, line 68: Here you are just testing if form data is valid. For correct testing you should add a bigger dataset for each model and test that search result Contains records that should be there and NotContains records that need to be filtered out

taxi/views.py
```diff
@@ -1,12 +1,19 @@
 from django.contrib.auth.decorators import login_required
+from django.contrib.auth.mixins import LoginRequiredMixin
 from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views import generic
-from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import (
```
Oleksl888: change relative import here

taxi/views.py
```diff
@@ -57,7 +79,22 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
     paginate_by = 5
-    queryset = Car.objects.select_related("manufacturer")
+
+    def get_context_data(self, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        model = self.request.GET.get("model", "")
+        context["car_search_form"] = CarsSearchForm(
```
Oleksl888: you could rename all the search forms with same name in context and put this form into includes in the templates so you don't have to repeat

Oleksl888 (changes requested):

Arsen-hrynevych (approved):
Overall ok

PR #670, opened 2024-09-16
first_try

yashkunn:


yashkunn:
image
image
image
image

taxi/views.py
```diff
@@ -36,6 +43,22 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
+        else:
+            return queryset
```
MNDonut: 1. call super method to get an initial queryset in get_queryset
2. for get_context_data do you need to specify class in case only one of the parent classes has this method?
fix everywhere
3. you don't need else statement because if statement has return. fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -36,6 +43,22 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
+        else:
```
Oleksl888: you don't need else here

taxi/views.py
```diff
@@ -57,7 +80,22 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        model = self.request.GET.get("model", "")
+        context["search_form"] = CarSearchForm(
+            initial={"model": model}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = super().get_queryset().prefetch_related("manufacturer")
```
Oleksl888: why use prefetch on m2o field?

taxi/views.py
```diff
@@ -57,7 +80,22 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+        form = CarSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(model__icontains=form.cleaned_data["model"])
+        else:
```
Oleksl888: else is redundant - fix everywhere

templates/taxi/driver_list.html
```diff
@@ -1,14 +1,19 @@
...
-        Create
-      </a>
-    </h1>
+  <h1>
+    Driver List
+    <a href="{% url 'taxi:driver-create' %}" class="btn btn-primary link-to-page">
+      Create
+    </a>
+  </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888: search form is repeated in every template. can you make it DRY?

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,84 @@
...
+        self.car = Car.objects.create(
+            model="Test Model",
+            manufacturer=self.manufacturer
+        )
+
+    def test_driver_search_form(self):
+        form_data = {"username": "testdriver"}
+        form = DriverSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        drivers = get_user_model().objects.filter(
```
Oleksl888, line 74: You cannot be sure that your search functionality works as intended if you test only on one record. Use bigger data sets  + you need to assert that correct results are present in response and incorrect results are not there.

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Nattalli, line 9: recommend you use absolute imports instead of relative

templates/includes/search_form.html
```diff
@@ -0,0 +1,5 @@
+{% load crispy_forms_filters %}
+<form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit" value="🔎">
+</form>
```
Nattalli, line 5: add empty line in the end of file

Nattalli (approved):
not bad but fix small comments

PR #671, opened 2024-09-17
Solution Politov

Alexandr-Politov:


Alexandr-Politov:
image
image

taxi/views.py
```diff
@@ -1,16 +1,20 @@
...
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (DriverCreationForm,
+                    DriverLicenseUpdateForm,
+                    CarForm,
+                    SearchForm)
```
MNDonut: fix imports

taxi/views.py
```diff
@@ -34,7 +38,20 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+    paginate_by = 2
+
+    def get_context_data(self, *, object_list=None, **kwargs) -> dict:
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        search = self.request.GET.get("search", "")
+        context["search_form"] = SearchForm(initial={"search": search})
+        return context
+
+    def get_queryset(self) -> QuerySet:
+        queryset = Manufacturer.objects.all()
```
MNDonut: call super method instead of hardcoding queryset

taxi/views.py
```diff
@@ -34,7 +38,20 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     model = Manufacturer
     context_object_name = "manufacturer_list"
     template_name = "taxi/manufacturer_list.html"
-    paginate_by = 5
+    paginate_by = 2
+
+    def get_context_data(self, *, object_list=None, **kwargs) -> dict:
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
```
MNDonut: Does LoginRequiredMixin has get_context_data so that you need to specify whose method you should call?

taxi/tests/tesf_forms.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+
+    def test_search_form(self) -> None:
+        form_data = {"username": "test"}
+        form = SearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        drivers = get_user_model().objects.filter(
+            username__icontains=form_data["username"]
+        )
+        self.assertEqual(list(drivers), [self.driver])
```
MNDonut: do you test search for cars and manufacturers?

MNDonut (changes requested):

taxi/views.py
```diff
@@ -85,6 +117,19 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        search = self.request.GET.get("search", "")
+        context["search_form"] = SearchForm(initial={"search": search})
+        return context
+
+    def get_queryset(self) -> QuerySet:
+        queryset = Driver.objects.all()
+        filter_ = self.request.GET.get("search")
+        if filter_:
+            return queryset.filter(username__icontains=filter_)
+        return queryset
```
fix everywhere

taxi/views.py
```diff
@@ -57,11 +76,24 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+        queryset = Car.objects.select_related("manufacturer")
+        filter_ = self.request.GET.get("search")
+        if filter_:
+            return queryset.filter(model__icontains=filter_)
+        return queryset
 
 
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
+    queryset = Car.objects.all()
```
MNDonut: no need to set it, because by default Django takes model.objects.all()

taxi/views.py
```diff
@@ -57,11 +76,24 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+        search = self.request.GET.get("search", "")
+        context["search_form"] = SearchForm(initial={"search": search})
+        return context
+
+    def get_queryset(self) -> QuerySet:
+        queryset = Car.objects.select_related("manufacturer")
+        filter_ = self.request.GET.get("search")
+        if filter_:
+            return queryset.filter(model__icontains=filter_)
+        return queryset
```
fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #672, opened 2024-09-17
implement search and tests

HalynaRohatska:
2024-09-17_21h11_21
2024-09-17_21h11_56
2024-09-17_21h12_14
2024-09-17_21h12_40
2024-09-17_21h12_53
2024-09-17_21h13_13


taxi/views.py
```diff
@@ -36,6 +42,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
```
MNDonut: call super method to get an initial queryset. fix everywhere

taxi/views.py
```diff
@@ -36,6 +42,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
```
MNDonut: you don't need to specify class and self(however it was required in 2 python) because Mixin doesn't have this method

taxi/tests/test_views.py
```diff
@@ -0,0 +1,105 @@
...
+DRIVER_URL = reverse("taxi:driver-list")
+
+
+class PublicManufacturerTest(TestCase):
+    def test_login_required(self):
+        res = self.client.get(MANUFACTURER_URL)
+        self.assertNotEqual(res.status_code, 200)
+
+
+class PrivetManufacturerTest(TestCase):
```
MNDonut, line 19: don't see search tests for all models

MNDonut (changes requested):

taxi/tests/test_views.py
```diff
@@ -0,0 +1,249 @@
...
+    def test_delete_driver(self):
+        response = self.client.post(
+            reverse("taxi:driver-delete", args=[self.driver_1.id])
+        )
+        self.assertEqual(response.status_code, 302)
+        self.assertFalse(Driver.objects.filter(id=self.driver_1.id).exists())
+
+
+class DriverSearchFormTest(TestCase):
+    def test_valid_data(self):
```
Oleksl888: Here you are just testing if form data is valid. For correct testing you should add a bigger dataset for each model and test that search result Contains records that should be there and NotContains records that need to be filtered out

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888: this can go in includes because it repeats in three templates

Oleksl888 (changes requested):

taxi/tests/test_views.py
```diff
@@ -0,0 +1,396 @@
...
+        self.car_2 = Car.objects.create(
+            model="Car 2",
+            manufacturer=self.manufacturer
+        )
+        self.car_3 = Car.objects.create(
+            model="Car 3",
+            manufacturer=self.manufacturer
+        )
+
+    def test_search_car_with_valid_data(self):
```
Oleksl888, line 231: Would make sense to add a case when you search for "Car 1" and then check if "Car 1" is in results and "Car 2 & 3" is not in results at the same time

Oleksl888 (approved):
Approved - please check last note

PR #673, opened 2024-09-17
Task Solution

achalaidiuk:


taxi/views.py
```diff
@@ -36,12 +44,38 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        name = self.request.GET.get("name")
+        if name:
+            return queryset.filter(name__icontains=name)
+        return queryset
```
fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #674, opened 2024-09-17
Develop

panichqa:


taxi/views.py
```diff
@@ -36,6 +43,23 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(
+                name__icontains=form.cleaned_data["name"]
+            )
+        return queryset
+
```
fix everywhere

taxi/views.py
```diff
@@ -59,6 +83,23 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(CarListView, self).get_context_data(**kwargs)
+        model = self.request.GET.get("model", "")
+        context["search_form"] = CarSearchForm(
+            initial={"model": model}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Car.objects.select_related("manufacturer")
```
MNDonut: you define queryset twice. as a class attribute and inside get_queryset method

MNDonut (changes requested):

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Nattalli: recommend you use absolute imports instead of relative

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,65 @@
...
+        form = CarSearchForm(data={"model": "Skyline"})
+        self.assertTrue(form.is_valid())
+
+    def test_manufacturer_search_form(self):
+        form = ManufacturerSearchForm(data={"name": "Nissan"})
+        self.assertTrue(form.is_valid())
+
+    def test_driver_search_form(self):
+        form = DriverSearchForm(data={"username": "driver1"})
+        self.assertTrue(form.is_valid())
```
Nattalli: make the tests of search feature more detailed

Nattalli (changes requested):

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,81 @@
...
+        form = CarForm(data=form_data)
+        self.assertTrue(form.is_valid())
+
+    def test_manufacturer_search_results(self):
+        form = ManufacturerSearchForm(data={"name": "Nissan"})
+        self.assertTrue(form.is_valid())
+        results = Manufacturer.objects.filter(
+            name__icontains=form.cleaned_data["name"]
+        )
+        self.assertIn(self.manufacturer, results)
```
Oleksl888, line 48: For correct testing you should add a bigger dataset for each model and test that search result Contains records that should be there and NotContains records that need to be filtered out

templates/taxi/car_list.html
```diff
@@ -7,15 +8,31 @@ <h1>
       Create
     </a>
   </h1>
+    <form method="get" action="" class="form-inline">
```
Oleksl888: this can go in `includes` because it repeats in three tempaltes

Oleksl888 (changes requested):

taxi/tests/test_forms.py
Oleksl888: For correct testing you should check in one test that it contains correct result and does not contain incorrect records. Also you still have only  one manufacturer created. Make at least three of each records (Manufacturer, Car, Driver) so we can be sure search works correctly

templates/includes/search_form.html
```diff
@@ -0,0 +1,7 @@
+{% load crispy_forms_filters %}
+<!-- templates/includes/search_form.html -->
+
+<form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
```
Oleksl888, line 6: check indentation - should be 2 spaces everywhere in html

Oleksl888 (changes requested):

MNDonut (changes requested):
Did you push your commits? :)

Arsen-hrynevych (approved):
Overall ok, but consider not violating DRY

PR #677, opened 2024-09-18
solution 1.0

Narberal90:


Narberal90:
Знімок екрана 2024-09-18 184200
Знімок екрана 2024-09-18 184054
Знімок екрана 2024-09-18 184037
Знімок екрана 2024-09-18 184019
Знімок екрана 2024-09-18 184005
Знімок екрана 2024-09-18 183942
Знімок екрана 2024-09-18 183926

taxi/views.py
```diff
@@ -85,10 +123,34 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+    def get_queryset(self):
+        queryset = Driver.objects.all()
+        form = DriverUsernameSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(
+                username__icontains=form.cleaned_data[
+                    "username"
+                ]
+            )
```
fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -169,13 +169,11 @@ class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
...
 def toggle_assign_to_car(request, pk):
-    driver = Driver.objects.get(id=request.user.id)
-    if (
-        Car.objects.get(id=pk) in driver.cars.all()
-    ):  # probably could check if car exists
-        driver.cars.remove(pk)
+    car = get_object_or_404(Car, id=pk)
+    driver = request.user
+    if car in driver.cars.all():
+        driver.cars.remove(car)
```
MNDonut: use .exists instead of "in" operator because for "in" you should fetch all records and loop through them. instead of it you can do it via SQL

MNDonut (changes requested):

templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888, line 11: this form repeats on 3 different templates - would make sense to make it an `include`

Oleksl888 (approved):
Well done!

PR #680, opened 2024-09-18
task done

Sergi0bbb:
driver list search:
image

car list search:
image

manufacturers list search:
image

### DIFF
--- a/taxi/tests/forms/test_search.py
+++ b/taxi/tests/forms/test_search.py
@@ -1,6 +1,12 @@
+from django.contrib.auth import get_user_model
 from django.test import TestCase
-
+from django.urls import reverse
 from taxi.forms import DriverSearchForm, CarSearchForm, ManufacturerSearchForm
+from taxi.models import Manufacturer, Car, Driver
+
+DRIVER_LIST_URL = reverse("taxi:driver-list")
+CAR_LIST_URL = reverse("taxi:car-list")
+MANUFACTURER_LIST_URL = reverse("taxi:manufacturer-list")
 
 
 class DriverSearchFormTest(TestCase):
@@ -8,6 +14,25 @@ class DriverSearchFormTest(TestCase):
         self.form_data = {
             "username": "bob"
         }
+        Driver.objects.create(
+            username="max",
+            password="testPassword123",
+            license_number="D2345670"
+        )
+        Driver.objects.create(
+            username="serg",
+            password="testPassword123",
+            license_number="D3345670"
+        )
+        Driver.objects.create(
+            username="jan",
+            password="testPassword123",
+            license_number="D4345670"
+        )
+        self.client.force_login(get_user_model().objects.create(
+            username="bob",
+            password="testPassword123"
+        ))
 
     def test_valid_form(self):
         form = DriverSearchForm(self.form_data)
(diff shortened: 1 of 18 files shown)


taxi/views.py
```diff
@@ -36,6 +43,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(
+            initial={"name": name}
+        )
+        return context
+
```
MNDonut: Do you need to define class in super when only one parent class has this method?
fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -59,6 +81,22 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        model = self.request.GET.get("model", "")
+        context["search_form"] = CarSearchForm(
+            initial={"model": model}
+        )
+        return context
+
+    def get_queryset(self):
+        form = CarSearchForm(self.request.GET)
+        if form.is_valid():
+            return self.queryset.filter(
```
MNDonut: it's not recommended to use self.queryset because if you inherit from another view, that has get_queryset() method overrided, but not queryset attribute, then you wan't reuse this logic. fix everywhere

MNDonut (changes requested):

taxi/views.py
```diff
@@ -36,6 +43,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        name = self.request.GET.get("name")
+        if name:
```
Oleksl888: missing the check that the form is valid

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Oleksl888: realtive import needs to be changed

taxi/views.py
```diff
@@ -85,6 +124,21 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        username = self.request.GET.get("username", "")
+        context["search_form"] = DriverSearchForm(
+            initial={"username": username}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        username = self.request.GET.get("username")
+        if username:
```
Oleksl888: here also - form validation is not performed

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
```
Oleksl888: same form repeats on three templates. Make it DRY

taxi/tests/forms/test_search.py
```diff
@@ -0,0 +1,75 @@
...
+        self.form_data = {
+            "name": "bob"
+        }
+
+    def test_valid_form(self):
+        form = ManufacturerSearchForm(self.form_data)
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data, self.form_data)
+
+    def test_valid_char_field_kwargs(self):
```
Oleksl888, line 160: these tests only check form validity but not the actual search results. 
You need to create a set of users/cars/manufacturers and  assert that correct results are present in response after search and incorrect results are not there.

Oleksl888 (changes requested):

Arsen-hrynevych (approved):
Overall ok

PR #681, opened 2024-09-19
Solution

mihavryliuk:


taxi/views.py
```diff
@@ -85,6 +121,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        return context
+
+    def get_queryset(self):
+        queryset = Driver.objects.all()
+        form = DriverSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(
+                username__icontains=form.cleaned_data["username"]
+            )
+        return queryset
```

MNDonut (changes requested):

taxi/views.py
```diff
@@ -57,7 +79,21 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+            initial={"model": model}
+        )
+        return context
+
+    def get_queryset(self):
+        queryset = Car.objects.select_related("manufacturer")
+        form = CarSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(model__icontains=form.cleaned_data["model"])
+        return queryset
```
2. Set this queryset as a class attribute

MNDonut (changes requested):

templates/taxi/car_list.html
```diff
@@ -7,16 +8,34 @@ <h1>
       Create
     </a>
   </h1>
+
+  <form method="get" action="" class="form-inline">
```
Oleksl888, line 12: this form is the same everywhere - it can go to includes

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,90 @@
...
+        }
+        form = DriverCreationForm(data=form_data)
+
+        if not form.is_valid():
+            print(form.errors)
+
+        self.assertTrue(form.is_valid())
+
+    def test_manufacturer_search_form(self):
+        form = ManufacturerSearchForm(data={"name": "Test"})
```
Oleksl888: This is not a precise implementation of search tests because you are adding only one record for each model. For correct testing you should add a bigger dataset for each model and that search result Contains records that should be there and NotContains records that need to be filtered out

Oleksl888: *Try to use fixture to upload big test data sets

Oleksl888 (changes requested):

templates/taxi/car_list.html
Oleksl888: This is not fixed - you can put this form in includes and reuse it on three list templates instead

Oleksl888 (approved):
Approved, but please check last comments to fix

PR #682, opened 2024-09-19
solution

mnezbrytska:


mnezbrytska:
Screenshot 2024-09-19 094208
Screenshot 2024-09-19 094202
Screenshot 2024-09-19 094134

taxi/views.py
```diff
@@ -36,6 +44,21 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        context["search_form"] = DriverSearchForm(
+            initial={"model": name}
+        )
+        return context
+
+    def get_queryset(self):
+        form = ManufacturerSearchForm(self.request.GET)
+        queryset = Manufacturer.objects.all()
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
```
MNDonut: call super method to get an initial queryset in get_queryset
for get_context_data do you need to specify class in case only one of the parent classes has this method?
fix everywhere

MNDonut (changes requested):

taxi/test/test_forms.py
```diff
@@ -0,0 +1,63 @@
...
+            name="test_manufacturer",
+            country="test_country"
+        )
+        self.car = Car.objects.create(
+            model="test_model",
+            manufacturer=self.manufacturer
+        )
+
+    def test_car_search_form_valid_data(self) -> None:
+        form_data = {"model": "test_model"}
```
Oleksl888, line 57: You cannot be sure that your search functionality works as intended if you test only on one record. Use bigger data sets + you need to assert that correct results are present in response and incorrect results are not there.

taxi/views.py
```diff
@@ -6,7 +8,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Oleksl888: change related import here

templates/taxi/car_list.html
```diff
@@ -7,6 +8,12 @@ <h1>
       Create
     </a>
   </h1>
+  
+  <form method="get" action="" class="form-inline">
```
Oleksl888: Same search form is repeated in every template. can you make it DRY?

Oleksl888 (changes requested):

taxi/test/test_forms.py
```diff
@@ -0,0 +1,74 @@
...
+        form = CarSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        cars = Car.objects.filter(
+            model__icontains=form_data["model"]
+        )
+        self.assertIn(self.car_1, cars)
+        self.assertIn(self.car_2, cars)
+        self.assertNotIn(self.non_matching_car, cars)
+        self.assertEqual(list(cars), [self.car_1, self.car_2])
```
Oleksl888, line 66: This is correct - but you also need to implement the same for drivers and manufacturers

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #683, opened 2024-09-19
Solution

pashawarganov:
Cars:
image

Drivers:
image

Manufacturers:
image


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,117 @@
...
+        self.assertEqual(
+            list(response.context["manufacturer_list"]),
+            list(self.manufacturers)
+        )
+
+    def test_driver_search_form_return_right_object(self):
+        response = self.client.get(DRIVER_URL, {"username": "bob1231"})
+        self.assertEqual(
+            list(response.context["driver_list"]),
+            [self.drivers[1]]
```
dimak20: Is it good to compare drivers by their order in the list? What would happen if you changed the order in self.drivers? I think it would be better to search by username and then make an assertion.

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+CAR_URL = reverse("taxi:car-list")
+
+
+class PublicManufacturerTest(TestCase):
+    def test_login_required(self):
+        response = self.client.get(MANUFACTURER_URL)
+        self.assertNotEqual(200, response.status_code)
+
+
+class PrivateManufacturerTest(TestCase):
```
dimak20, line 18: Your tests are quite comprehensive, but I think It'd be better to test the search forms in the test_views.py using urls with such params. Try adding "?param=query" to your url in order to check whether search is correct)

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        Manufacturer.objects.create(name="Audi", country="Germany")
+        Manufacturer.objects.create(name="Ford", country="USA")
+        self.manufacturers = Manufacturer.objects.all()
+        self.client.force_login(self.user)
+
+    def test_retrieve_manufacturer(self):
+        response = self.client.get(MANUFACTURER_URL)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
```
dimak20, line 32: Good, but you have pagination in your views. You could also try searching on different pages to check whether pagination is correct. There is also simmilar pattern ("?page={some page}")

taxi/tests/test_forms.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for license_number in license_numbers:
+            form_data["license_number"] = license_number
+            form = DriverCreationForm(data=form_data)
+            self.assertFalse(form.is_valid())
+
+
+class SearchFormsTests(TestCase):
+    def setUp(self):
+        self.manufacturers = Manufacturer.objects.bulk_create([
```
dimak20: I think It'd be more readable to start the bracket on a new line

taxi/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+    def setUp(self):
+        self.user = get_user_model().objects.create_user(
+            username="bob123",
+            password="test123",
+        )
+        self.manufacturer = Manufacturer.objects.create(
+            name="Ford",
+            country="USA"
+        )
+        self.cars = Car.objects.bulk_create([
```
dimak20, line 83: I think It'd be more readable to start the bracket on a new line

taxi/views.py
```diff
@@ -33,8 +40,24 @@ def index(request):
 class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     model = Manufacturer
     context_object_name = "manufacturer_list"
-    template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
+    queryset = Manufacturer.objects.all()
```
dimak20: Do you need to define queryset in the class when you override the get_queryset method?

dimak20 (commented):
Overall good job :)

dimak20 (approved):

MasakDirt:
Well, I do not know why I also need to review your PR, Dima already reviewed it, so good job!)

MasakDirt (approved):

taxi/views.py
MNDonut: indeed, if you need access the default queryset, you can call super().get_queryset inside get_queryset

taxi/tests/test_forms.py
MNDonut: yeah, agree. It's better either to make some method for search or create a dict, to list, to explicitly pass the key like A1, so that it'll be easier for other devs to understand this logic

MNDonut (changes requested):

taxi/views.py
```diff
@@ -85,6 +124,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        username = self.request.GET.get("username", "")
+        context["search_form"] = DriverSearchForm(
+            initial={"username": username}
+        )
+        return context
+
+    def get_queryset(self):
+        self.queryset = Driver.objects.prefetch_related("cars")
```
MNDonut: It isn't recommended to do it (self.queryset = ...). get_queryset() should return queryset, but don't set it

You see, when django needs queryset, it calls get_queryset() method because you may need dynamic queryset depending on some conditions. Tis method should return queryset. In case this method is not overrided, django'll take queryset from queryset attribute

taxi/views.py
```diff
@@ -33,9 +40,25 @@ def index(request):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(
+            initial={"name": name}
+        )
+        return context
+
+    def get_queryset(self):
+        self.queryset = Manufacturer.objects.all()
```
MNDonut: don't set queryset inside this method. only return it

taxi/views.py
```diff
@@ -57,7 +80,23 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        model = self.request.GET.get("model", "")
+        context["search_form"] = CarSearchForm(
+            initial={"model": model}
+        )
+        return context
+
+    def get_queryset(self):
+        self.queryset = Car.objects.select_related("manufacturer")
```
MNDonut: fix it

MNDonut (changes requested):

Oleksl888 (approved):
Approved! :muscle:

PR #684, opened 2024-09-19
'Solution'

ABaidin:
Screenshot 2024-09-19 174805
Screenshot 2024-09-19 174816
Screenshot 2024-09-19 174845


taxi/views.py
```diff
@@ -59,6 +80,20 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
     queryset = Car.objects.select_related("manufacturer")
```
pashawarganov, line 81: Do you need to define queryset in the class when you override the get_queryset method?

pashawarganov (approved):
Good job!

taxi/views.py
```diff
@@ -36,6 +43,20 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        context["name"] = name
+        context["search_form"] = ManufacturerSearchForm(initial={"name": name})
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
+        return queryset
```
fix everywhere

MNDonut (changes requested):

tests/test_views.py
```diff
@@ -0,0 +1,472 @@
...
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "taxi/car_confirm_delete.html")
+
+    def test_car_delete(self):
+        self.client.post(reverse("taxi:car-delete", args=[self.car.id]))
+        cars = Car.objects.filter(id=self.car.id)
+        self.assertEqual(len(cars), 0)
+
+    # I'm tired, boss
```
Oleksl888: :crying_cat_face:

tests/test_views.py
```diff
@@ -0,0 +1,472 @@
...
+        self.assertEqual(
+            len(response.context["driver_list"]), self.paginated_by
+        )
+
+    def test_driver_list_search(self):
+        response = self.client.get(self.driver_list_url + "?username=7")
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context["driver_list"]), 1)
+        self.assertContains(response, "TestDriver7")
+        self.assertNotContains(response, "TestDriver8")
```
Oleksl888: Are we missing something? Maybe you forgot to push fixture with test data? Right now this assert will always be True - because TestDriver8 was never added. I see the same in other tests as well.
Also this file is too big - would make sense to break down in smaller files

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #689, opened 2024-09-22
Resolve

ihorutkin:
image
image
image
image
image


taxi/views.py
```diff
@@ -31,11 +39,26 @@ def index(request):
...
-    template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_queryset(self):
+        name = self.request.GET.get("name")
+        queryset = Manufacturer.objects.all()
+
+        if name:
+            return queryset.filter(name__icontains=name)
+        return queryset
```
BastovOleksandr, line 51: It would be more logical to use "name__startswith=name" lookup here
If I get this homework rejected, I will change it for myself too :)
Just telling

taxi/test/test_views.py
```diff
@@ -0,0 +1,127 @@
...
+    def setUp(self):
+        self.user = get_user_model().objects.create(
+            username="new_driver",
+            password="root1234",
+        )
+        self.client.force_login(self.user)
+
+    def test_index_counters(self):
+        response = self.client.get(reverse("taxi:index"))
+        self.assertEqual(response.context["num_cars"], 10)
```
BastovOleksandr, line 36: usually add a blank line between the input and the check
line 35, then empty one

taxi/test/test_views.py
```diff
@@ -0,0 +1,127 @@
...
+
+            Driver.objects.create_user(
+                username=f"driver {id_}",
+                license_number=str(id_)
+            )
+
+            Car.objects.create(
+                model=f"Model {id_}",
+                manufacturer=manufacturer
+            )
```
BastovOleksandr, line 58: Perhaps it makes sense to move the general code into a separate function and call it in setUpTestData methods?
check, it is the same in ManufacturerTest and IndexTest

BastovOleksandr (approved):
To be honest, there are probably many places where you can find fault with the tests. But for this you need: 1) much more time than 1 hour for checking. 2) how to write tests correctly is a rather controversial issue.

ihorutkin:
Hi, thanks for you review, i will change it in my project


вс, 22 сент. 2024 г. в 22:43, Bastov Oleksandr ***@***.***>:

.github/workflows/test.yml
```diff
@@ -27,6 +27,6 @@ jobs:
       - name: Run flake8
         run: flake8
 
-      - name: Run tests
+      - name: Run test
```
Oleksl888, line 30: why would you change a config file instead of simply naming your test directory `tests`?

taxi/views.py
Oleksl888: I think icontains is good - we ust looking for any match in name- not only from the beginning

taxi/views.py
```diff
@@ -31,11 +39,26 @@ def index(request):
...
 
 class ManufacturerListView(LoginRequiredMixin, generic.ListView):
-    model = Manufacturer
-    context_object_name = "manufacturer_list"
-    template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_queryset(self):
+        name = self.request.GET.get("name")
+        queryset = Manufacturer.objects.all()
```
Oleksl888: Here would make more sense to get a queryset from super()

taxi/views.py
```diff
@@ -6,7 +7,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Oleksl888: fix relative import here please

taxi/views.py
```diff
@@ -31,11 +39,26 @@ def index(request):
...
+    def get_queryset(self):
+        name = self.request.GET.get("name")
+        queryset = Manufacturer.objects.all()
+
+        if name:
+            return queryset.filter(name__icontains=name)
+        return queryset
+
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data()
```
Oleksl888: you don't need to pass args super explicitly here, `.get_context_data()` should be called with kw arguments

Oleksl888: Please fix the same in other views as well

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
```
Oleksl888: this form repeats in 3 templates, can you optimize?

taxi/test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+            "password2": "Root1234"
+        }
+        form = DriverCreationForm(form_data)
+        self.assertTrue(form.is_valid())
+
+    def test_driver_search_form_with_arg(self):
+        form_data = {
+            "username": "new_driver",
+        }
+        form = DriverUsernameSearchForm(data=form_data)
```
Oleksl888, line 171: Search test functionality must be done on bigger data sets + you need to assert that correct results are present in response and incorrect results are not there. Right now testing if the form is valid is only checking if search query has been passed to the form

taxi/test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+        form_data = {
+            "username": "",
+        }
+        form = DriverUsernameSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+
+    def test_update_license_form_correct_num(self):
+        form_data = {"license_number": "HRN84739"}
+        form = DriverLicenseUpdateForm(form_data)
+        self.assertTrue(form.is_valid())
```
Oleksl888, line 192: this only checks how the validators work but doesn't test that the license number was actually updated

Oleksl888 (changes requested):

.github/workflows/test.yml
ihorutkin: It was casualy autorefactor

taxi/test/test_forms.py
Oleksl888: This is not resolved - you need to test search functionality properly. Add more data and check if correct data appears in result and incorrect doesn't appear at the same time

Oleksl888 (changes requested):

taxi/test/test_forms.py
```diff
@@ -0,0 +1,140 @@
...
+        form_data = {
+            "username": "",
+        }
+        form = DriverUsernameSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+        search_username = form.cleaned_data.get("username")
+        results = get_user_model().objects.filter(
+            username__icontains=search_username
+        )
+        self.assertEqual(results.count(), 3)
```
Oleksl888, line 184: example of correct implementation of search tests would be: your search form has 'new_driver',  then you check in results if new_driver is in results and other drivers are not in result. and you need to do the same for Cars and Manufacturers

taxi/views.py
```diff
@@ -32,10 +40,26 @@ def index(request):
...
+
+        if name:
+            return queryset.filter(name__icontains=name)
+        return queryset
+
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        name = self.request.GET.get("name", "")
+
```
Oleksl888, line 57: shouldn't you pas the `name` to the form and check if its valid?

Oleksl888 (changes requested):

taxi/views.py
Oleksl888: sorry this should be done in get_queryset method

MNDonut (approved):

PR #690, opened 2024-09-23
first

egor250330:


egor250330:
car_list
create_driver
drivers_list
manufacturer_list
taxi_home

test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+            username="new_driver",
+            password="root1234",
+        )
+        self.client.force_login(self.user)
+
+        self.manufacturer = Manufacturer.objects.create(
+            name="Test manufacture", country="USA"
+        )
+
+    def test_driver_form(self):
```
ihorutkin: Make name of you tests functions more sense (For example `test_driver_creation_form_is_valid_with_valid_data`)

test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+    def setUp(self):
+        self.user = get_user_model().objects.create_user(
+            username="new_driver",
+            password="root1234",
+        )
+        self.client.force_login(self.user)
+
+        self.manufacturer = Manufacturer.objects.create(
+            name="Test manufacture", country="USA"
+        )
```
ihorutkin: Each test class (`TestCarForms`, `TestManufacturerForms`, `TestDriverForms`) contains the same code to create a user and a manufacturer. You might consider moving common code into a base class to avoid duplication. For example, create a base test class with the setUp method, which will be inherited by all other test classes

test/test_views.py
```diff
@@ -0,0 +1,85 @@
...
+        self.car = Car.objects.create(
+            model="M5",
+            manufacturer=self.manufacturer
+        )
+
+        response = self.client.delete(
+            reverse("taxi:car-delete", args=[self.car.id])
+        )
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(Car.objects.last(), None)
```
ihorutkin: The `test_car_delete` test queries to delete a car, but then checks that the last car in the database is None. This may cause an error if there are other cars in the database. Instead, you can verify that a specific car was deleted:

ihorutkin (approved):

taxi/views.py
```diff
@@ -6,7 +7,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
```
Oleksl888: change relative imports here please

taxi/views.py
```diff
@@ -36,6 +44,24 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_queryset(self):
+        name = self.request.GET.get("name")
+        queryset = Manufacturer.objects.all()
+
+        if name:
```
Oleksl888: name should be passed to form to ensure validation is in place

taxi/views.py
```diff
@@ -36,6 +44,24 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+            return queryset.filter(name__icontains=name)
+        return queryset
+
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super(ManufacturerListView, self).get_context_data()
```
Oleksl888: no need to pass class name to super() explicitly; 
**kwargs should be passed to get_context_data

templates/taxi/car_list.html
```diff
@@ -7,6 +8,11 @@ <h1>
       Create
     </a>
   </h1>
+  <form action="" method="get" class="form-inline">
```
Oleksl888: same form is added on all three templates - can you refactor to avoid repetition?

test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+        self.manufacturer = Manufacturer.objects.create(
+            name="Manufacture Test", country="China"
+        )
+
+    def test_manufacturer_search_form_with_arg(self):
+        form_data = {
+            "name": "Manufacture Test",
+        }
+        form = ManufacturerNameSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
```
Oleksl888: not only you should test that the form is valid - you should check if search prompt produced correct results. Assert if correct manufacturers are in returned data and incorrect manufacturers are not in data at the same time

test/test_forms.py
```diff
@@ -0,0 +1,112 @@
...
+        form_data = {
+            "username": "",
+        }
+        form = DriverUsernameSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+
+    def test_update_license_form_correct_num(self):
+        form_data = {"license_number": "HRN84739"}
+        form = DriverLicenseUpdateForm(form_data)
+        self.assertTrue(form.is_valid())
```
Oleksl888: this only checks that form is valid - you also need to check if record was actually updated

test/__init__.py
Oleksl888, line 1: tests should be inside taxi app

Oleksl888 (changes requested):

taxi/test/test_forms.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+
+    def test_manufacturer_search_form_with_arg(self):
+        form_data = {
+            "name": "Test",
+        }
+        form = ManufacturerNameSearchForm(data=form_data)
+        self.assertTrue(form.is_valid())
+
+        manufacturers = Manufacturer.objects.filter(
```
Oleksl888: to correctly test search functionality you need to make a request to client with search queryand then check if results are correct:
```
response = self.client.get(url, {"search": "search_string"})
self.assertEquals(response.status_code, 200)
self.assertContains(response, "search_string")
self.assertNotContains(response, "another_string")

File: taxi_service/settings.py [hunk 11]
```diff
@@ -91,26 +90,25 @@
...
 
-
 # Password validation
 # https://docs.djangoproject.com/en/4.0/ref/settings/#auth-password-validators
 
 AUTH_PASSWORD_VALIDATORS = [
     {
         "NAME": "django.contrib.auth.password_validation."
-        "UserAttributeSimilarityValidator",
+                "UserAttributeSimilarityValidator",
```
Oleksl888: fix indentation here

templates/taxi/car_list.html
Oleksl888, line 1: templates must be inside taxi app:
taxi/templates

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #698, opened 2024-10-05
'Solution'

Sonemon:
manu
drivers
cars
search
tests

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -40,8 +40,6 @@ class SearchMixin:
                     **{f"{self.model_search_field}__icontains": query}
                 )
                 return filtered_queryset
-                # return queryset.filter(
-                # **{f"{self.model_search_field}__icontains": query})
         return queryset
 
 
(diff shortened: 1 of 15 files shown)


templates/taxi/car_list.html
```diff
@@ -7,6 +7,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+    {{ search_form }}
+    <input type="submit" value="Search" class="btn btn-secondary">
+  </form>
   {% if car_list %}
```
vsmutok: Define separate template for search and include it into places you need

taxi/views.py
```diff
@@ -54,10 +92,29 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    #     )
+    #     return context
+    #
+    # def get_queryset(self):
+    #     queryset = self.queryset
+    #     form = SearchForm(self.request.GET, placeholder="Search by model")
+    #     if form.is_valid():
+    #         return queryset.filter(
+    #         model__icontains=form.cleaned_data["query"])
+    #     return queryset
```
vsmutok: Why you commented this?

taxi/views.py
```diff
@@ -6,7 +6,43 @@
...
+        form = self.get_search_form()
+        if form.is_valid():
+            query = form.cleaned_data.get("query", "")
+            if query:
+                filtered_queryset = queryset.filter(
+                    **{f"{self.model_search_field}__icontains": query}
+                )
+                return filtered_queryset
+                # return queryset.filter(
+                # **{f"{self.model_search_field}__icontains": query})
```
vsmutok: remove this comment

vsmutok (changes requested):

taxi/views.py
Sonemon: Because I originally did the tutorial, but then I saw the duplication of code and it became interesting to try to make a generic search mixin that would be universal across different models. So I made it, but I left a comment for myself, so that if the mixin doesn't fit, I can quickly replace it.

taxi/views.py
Sonemon: Okay, but why are my comments an error? I make them for me.

vasylhnatiuk (approved):

PR #732, opened 2024-11-20
solution

dmytrik:
Screenshot 2024-11-20 at 16 30 48
Screenshot 2024-11-20 at 16 30 58
Screenshot 2024-11-20 at 16 31 04
Screenshot 2024-11-20 at 16 31 10
Screenshot 2024-11-20 at 16 31 27
Screenshot 2024-11-20 at 16 31 32

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ flake8-variables-names==0.0.5
 pep8-naming==0.13.2
 django-debug-toolbar==3.2.4
 django-crispy-forms==1.14.0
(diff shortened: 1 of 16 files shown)


AnyoneClown (changes requested):
The tests are well written, but there are no cases with invalid data. It is worth writing them down as well.

Add a few test cases where the data will be intentionally incorrect (expect, for example, a status code with errors, etc.)

Arsen-hrynevych (approved):
ok

PR #745, opened 2024-11-26
solution

dxrrkwm:


taxi/tests.py
```diff
@@ -1,3 +1,32 @@
...
-# Create your tests here.
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+from django.urls import reverse
+
+from taxi.models import Manufacturer, Driver, Car
+
+
+class SearchTests(TestCase):
+    def setUp(self):
```
pohgen: you need to write tests for whole project not only for searching

taxi/urls.py
```diff
@@ -1,75 +1,79 @@
...
+    ),
+    path(
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete",
+    ),
+    path("drivers/", DriverListView.as_view(), name="search-drivers"),
+    path("cars/", CarListView.as_view(), name="search-cars"),
+    path("manufacturers/",
+         ManufacturerListView.as_view(), name="search-manufacturers"),
```
pohgen, line 76: you need to make search as a form, not as a path

pohgen (changes requested):
Good searching tests, but that's not all that you need to implement

PR #781, opened 2025-01-02
Solution

serg121192:


templates/taxi/car_list.html
```diff
@@ -1,12 +1,18 @@
...
     Car list
     <a href="{% url 'taxi:car-create' %}" class="btn btn-primary link-to-page">
       Create
     </a>
   </h1>
+  <br>
+  <form action="" method="get">
+    {{ search_form|crispy }}
+    <input type="submit" value="Search" class="btn btn-secondary">
+  </form>
```
pavlopro, line 14: create a separate template for search and include it wherever you need

pavlopro (changes requested):

PR #802, opened 2025-01-26
Solution

yablonskyiv:

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -5,8 +5,8 @@ from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
-from .models import Driver, Car, Manufacturer
-from .forms import (
+from taxi.models import Driver, Car, Manufacturer
+from taxi.forms import (
     DriverCreationForm,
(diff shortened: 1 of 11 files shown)


templates/taxi/car_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit" value="🔍">
+  </form>
```
vsmutok: Create a separate template for search form and include it where you need to avoid code duplication

vsmutok (changes requested):

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
```
SerhiyTo: Use absolute imports instead of relative imports (`from app.models import` instead of `from .models import`)

SerhiyTo (changes requested):

vsmutok (approved):
Looks Good!

PR #826, opened 2025-01-31
solution

yelizaveta-zh:


Nattalli (changes requested):
1) piplines are failed, fix them
2) add tests fot `post` methods, not `get` only

Nattalli (approved):

PR #834, opened 2025-02-03
Solution

Professor-Douel:


taxi/tests.py
```diff
@@ -1,3 +1,65 @@
...
+        response = self.client.get(reverse("taxi:driver-list"))
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(self.driver, response.context["object_list"])
+
+    def test_toggle_assign_to_car(self):
+        response = self.client.post(reverse(
+            "taxi:toggle-assign-to-car",
+            args=[self.car.id])
+        )
+        self.assertEqual(response.status_code, 302)
```
Nattalli: about `test_toggle_assign_to_car` -> you can also add checking before request that user. is assigned to the car and check again after request that user is not assigned to the car. That would make moresense then just having those tests

taxi/views.py
```diff
@@ -6,7 +6,8 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm, \
+    ManufacturerSearchForm, CarSearchForm, DriverSearchForm
```
Nattalli: better use absolute imports instead of relative

Nattalli (changes requested):

Nattalli (approved):

PR #842, opened 2025-02-27
Solution

juliazelena:


juliazelena:
Знімок екрана 2025-02-27 о 18 24 21
Знімок екрана 2025-02-27 о 18 24 07
Знімок екрана 2025-02-27 о 18 24 01
Знімок екрана 2025-02-27 о 18 23 46
Знімок екрана 2025-02-27 о 18 23 37

templates/taxi/car_list.html
```diff
@@ -7,6 +7,9 @@ <h1>
       Create
     </a>
   </h1>
+  {% block search %}
+    {% include "includes/search_form.html" %}
+  {% endblock %}
```
vsmutok, line 12: Create a separate template for search form and include it where you need to avoid code duplication

taxi/views.py
```diff
@@ -6,7 +6,8 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm, \
+    DriverSearchForm, CarSearchForm, ManufacturerSearchForm
```
vsmutok: better to use `()` then `\`

also It is better to use absolute imports rather than relative ones.

vsmutok (changes requested):

templates/taxi/car_list.html
juliazelena: What do you mean? This is a separate template search_form.html, which I used in templates for searching

SerhiyTo (approved):

PR #851, opened 2025-03-10
Solution

JohnyJDX:


taxi/forms.py
```diff
@@ -2,8 +2,22 @@
...
 from django.core.exceptions import ValidationError
+from taxi.models import Car, Driver, Manufacturer
 
-from taxi.models import Car, Driver
+
+class ManufacturerSearchForm(forms.ModelForm):
+    name = forms.CharField(
+        max_length=255,
+        required=False,
+        label="",
```
Nattalli, line 12: probably you can just remove that attribute

taxi/forms.py
```diff
@@ -17,6 +31,15 @@ class Meta:
         fields = "__all__"
 
 
+class CarSearchForm(forms.Form):
+    model = forms.CharField(
+        max_length=255,
+        required=False,
+        label="",
```
Nattalli, line 38: the same here

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Nattalli, line 9: suggest you use absolute imports, not relative

taxi/views.py
```diff
@@ -36,6 +43,19 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     template_name = "taxi/manufacturer_list.html"
     paginate_by = 5
 
+    def get_context_data(self, *, object_list=None, **kwargs):
```
Nattalli, line 46: If you don't use some parameter then better use `_` instead of `*`

```suggestion
    def get_context_data(self, _, object_list=None, **kwargs):
```

Nattalli: or if there a lot of parameters than just use `*args` even if you don't use that

Nattalli (changes requested):

PR #893, opened 2025-04-11
solution

HappyDen08:
photo_5460643956284061345_w
photo_5460643956284061338_w


AndyStarGitHub (approved):
Well done! I'd just mention one thing that I noticed. Probably I just can not find it. Will filtering work properly with pagination? Perhaps we should add something like that to pagination.html file:         <a href="?{% query_transform request page=page_obj.previous_page_number %}" class="page-link">prev</a>

PR #895, opened 2025-04-11
'Solution'

AndyStarGitHub:


Anollinna (approved):
I see you’ve separated tests into different files, which is a good practice. The code looks clean. Great job!
The only thing missing is a project presentation image — it would be a nice touch!

PR #900, opened 2025-04-12
Solution

d-stegantsev:
image
image
image


taxi/forms.py
```diff
@@ -50,3 +50,57 @@ def validate_license_number(
...
+#         widget=forms.TextInput(attrs={"placeholder": "Search by model"}),
+#     )
+#
+# class ManufacturerSearchForm(forms.Form):
+#     name = forms.CharField(
+#         max_length=63,
+#         required=False,
+#         label="",
+#         widget=forms.TextInput(attrs={"placeholder": "Search by name"}),
+#     )
```
Boyok1337: Delete the commented code.
Tip: If you want to keep the commented code in the future, provide information about why you want to keep it.

taxi/forms.py
```diff
@@ -50,3 +50,57 @@ def validate_license_number(
...
+    placeholder = "Search"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields[self.search_field_name] = forms.CharField(
+            max_length=63,
+            required=False,
+            label="",
+            widget=forms.TextInput(attrs={"placeholder": self.placeholder}),
+        )
```
Boyok1337, line 66: Good one!

taxi/views.py
```diff
@@ -6,7 +6,14 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 
 from .models import Driver, Car, Manufacturer
-from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
+from .forms import (
```
Boyok1337: Absolute imports better than relative

taxi/views.py
```diff
@@ -36,6 +43,19 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
+        context = super(ManufacturerListView, self).get_context_data(**kwargs)
+        name = self.request.GET.get("name", "")
+        context["search_form"] = ManufacturerSearchForm(initial={"name": name})
+        return context
+
+    def get_queryset(self):
+        queryset = Manufacturer.objects.all()
+        form = ManufacturerSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(name__icontains=form.cleaned_data["name"])
```
Boyok1337: Here you can only filter qs, then you can easily to add more conditions

taxi/views.py
```diff
@@ -59,6 +79,20 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        context["search_form"] = CarSearchForm(initial={"model": model})
+        return context
+
+    def get_queryset(self):
+        form = CarSearchForm(self.request.GET)
+        if form.is_valid():
+            return self.queryset.filter(
+                model__icontains=form.cleaned_data["model"]
+            )
+        return self.queryset
```
Boyok1337: The same

taxi/views.py
```diff
@@ -85,6 +119,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        return context
+
+    def get_queryset(self):
+        queryset = Driver.objects.all()
+        form = DriverSearchForm(self.request.GET)
+        if form.is_valid():
+            return queryset.filter(
+                username__icontains=form.cleaned_data["username"]
+            )
+        return queryset
```
Boyok1337: The same

Boyok1337 (changes requested):

taxi/forms.py
d-stegantsev: I refactored search forms to use BaseSearchForm but forgot to remove the old ones ))

Boyok1337 (approved):
If you override the get_queryset method, then there is no point in specifying the default

PR #930, opened 2025-05-06
Develop

olek007:
image
image
image


olek007:
@mateacademy-ai-mentor 
in tests fields driver.first_name and driver.last_name are set in setup.
Everything work correctly

PR #941, opened 2025-05-17
Solution

shinodegu:


shinodegu:
2.Driver List Pagination Test: The test assumes 5 drivers per page, but this must match the actual pagination setting in your view. Please verify your pagination configuration and update the test assertions accordingly to avoid false positives or negatives.
-You didn't take into account that I create a user before the loop and on the site each user is a driver. so I create one user manually and 5 through the loop. If I created only 5 users I wouldn't be able to run the tests.
4.User Profile Link in Sidebar: In your sidebar template, you are using {{ user.get_absolute_url }} in the href attribute. This does not call the method and will not generate the correct URL. You should use {{ user.get_absolute_url() }} (if it's a method) or {{ user.get_absolute_url }} (if it's a property), but most likely it needs to be called as a method.
Это так не работает и вызывает ошибку, ссылка перестает работает на сайте тоже.

Снимок экрана (70)
Остальное я исправил.

PR #1008, opened 2025-08-18
Implemented search for all list templates, wrote tests for models and…

Otst0y:
… forms.


Otst0y:
image
image
image

reinvvay (changes requested):
Platform gave me unfinished solution, so i have to decline it, as i see, you need to delete your requirements.txt file, in interpretator settings update every library, then in terminal: pip freeze > requirements.txt. This should help, as i understand you have compatibility problems with crispy and django (for example django version is not supported with current crispy/bootstrap version, so just update everything to newest version).

taxi/views.py
```diff
@@ -56,9 +82,27 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
-    paginate_by = 5
+    paginate_by = 2
     queryset = Car.objects.select_related("manufacturer")
 
+    def get_context_data(
+            self, *, object_list=..., **kwargs
```
L1nk3rrr, line 89: object_list? why you need it here

taxi/views.py
```diff
@@ -83,7 +127,25 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
 
 class DriverListView(LoginRequiredMixin, generic.ListView):
     model = Driver
-    paginate_by = 5
+    paginate_by = 2
+
+    def get_context_data(
+            self, *, object_list=..., **kwargs
```
L1nk3rrr, line 133: same

L1nk3rrr (approved):
Well done!

PR #1014, opened 2025-08-19
solution

Sashoxxx:
Знімок екрана 2025-08-19 о 14 00 15
Знімок екрана 2025-08-19 о 14 00 23
Знімок екрана 2025-08-19 о 14 00 29


taxi/tests/test_forms.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_manufacturer_search_form_valid_data(self):
+        form = ManufacturerSearchForm(data={"name": "Tesla"})
+        self.assertTrue(form.is_valid())
+        self.assertEqual(form.cleaned_data["name"], "Tesla")
+
+    def test_search_forms_allow_empty_data(self):
+        self.assertTrue(CarSearchForm(data={}).is_valid())
+        self.assertTrue(DriverSearchForm(data={}).is_valid())
+        self.assertTrue(ManufacturerSearchForm(data={}).is_valid())
```
yagginov: I think this is unnecessary, because here you are checking what Django guarantees us, specifically in the search forms we have not added anything new that would be worth checking.

yagginov (approved):
As Сергій said, "there are never too many tests," but if I were you, I would simply be too lazy to write so many tests.

PR #1019, opened 2025-08-19
sol

zqpiq:
Снимок экрана 2025-08-19 в 20 53 51
Снимок экрана 2025-08-19 в 20 54 01
Снимок экрана 2025-08-19 в 20 54 05
Снимок экрана 2025-08-19 в 20 54 09


Plovuezz (approved):
Nice, but i think you had to do also pagination with your list of objects after search, but everything else looks good. Good Job Tamerlan

PR #1111, opened 2025-12-16
Create search and tests

SofaD8:


taxi/tests.py
VadymVynokur, line 1: create multiple test files would be better imo

VadymVynokur (approved):
Good Job

PR #1122, opened 2025-12-18
Solution

oleksii-shcherbak-code:


tigersaymon (approved):

tigersaymon:
Some tests have failed. Don't forget to fix them. In general, it's better to create different test files for different models

PR #1300, opened 2026-07-30
Complete

Rostyk-D:
image
image
image
image
image


taxi/views.py
```diff
@@ -36,6 +43,28 @@ class ManufacturerListView(LoginRequiredMixin, generic.ListView):
...
     paginate_by = 5
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["search_form"] = ManufacturersNameSearchForm(
+            self.request.GET
+        )
+        return context
+
+    def get_queryset(self):
```
Sn1kls: This search pattern (`get_context_data` + `get_queryset`, building the same form twice) is repeated identically across all three list views. Consider extracting a small `SearchMixin` parametrised by `search_form_class` + `search_field`, so each view only declares those two attributes. Bonus: the form gets instantiated once per request instead of twice

taxi/views.py
```diff
@@ -111,7 +184,7 @@ class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
 def toggle_assign_to_car(request, pk):
     driver = Driver.objects.get(id=request.user.id)
     if (
-        Car.objects.get(id=pk) in driver.cars.all()
+            Car.objects.get(id=pk) in driver.cars.all()
```
Sn1kls: Use `get_object_or_404(Car, id=pk)` here so a bad pk returns 404 instead of raising `DoesNotExist`

taxi/forms.py
```diff
@@ -50,3 +50,30 @@ def validate_license_number(
         raise ValidationError("Last 5 characters should be digits")
 
     return license_number
+
+
+class CarModelSearchForm(forms.Form):
+    model = forms.CharField(
+        max_length=255,
+        required=False,
+        label=" ",
```
Sn1kls: Minor: an empty-string` label=""` is cleaner than a space here (same on the other two search forms)

Sn1kls (changes requested):

hladiivolodymyr (approved):

PR #1311, opened 2026-08-09
solution

mrkanis82-gif:


templates/taxi/manufacturer_list.html
```diff
@@ -7,6 +8,10 @@ <h1>
       Create
     </a>
   </h1>
+  <form method="get" action="" class="form-inline">
+    {{ search_form|crispy }}
+    <input class="btn btn-secondary" type="submit">
+  </form>
```
vsmutok: I recommend moving this piece of code into a separate module and then importing it into all the files where you need it to avoid duplication.

vsmutok (changes requested):

templates/taxi/manufacturer_list.html
mrkanis82-gif: Yeah, sounds logical, done.

hladiivolodymyr (approved):

